PR #94, opened 2020-10-21
Add documentations and movie related additions

eywalker:
* Add documentations to some utility functions found in `training.py` and also refined implementations.
* Notably, the default behavior of `TimeObjectiveTracker` is now **different** - to obtain the previous behavior, `add_creation_event=True` can be used. This was changed given that no one seems to be using this and also that it doesn't make much sense to relativize entry times against the creation time for the class object.
* Movie related additions in the form of file tree datasets and new transformer for stacking behavioral inputs have been adeed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Ignore the old package name
+mlutils
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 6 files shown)


KonstantinWilleke:
Would it be okay if you tag more people as reviewers? I'll take care of my PR and Claudius', so I won't be able to look at this completely.

ArneNx:
Could you please separate the changes that involve the movie data into a separate PR?

eywalker:
Separating movie related changes won't be trivial as the commits are intertwined. Changes even with respect to movie stuff is essentially a copy of existing approaches so I hope you can just review all together. As this is actively worked on, it's best for this to be reviewed soon or otherwise more changes will pile up.

KonstantinWilleke (approved):

PR #89, opened 2020-09-15
Add transforms and other features

KonstantinWilleke:
PR that has accumulated a bunch of small features.

- Scale transformation for the NeuroNormalizer in transforms.py
- solving a bug when normalizing the behavior states, and giving the eye_position the right name (`pupil_center`)
- adding a 5x5 laplace filter for the Laplace regularizer
- initializing a FileTree Dataset without behavior
- adding a new regularizer to the gaussian readout, that just returns the standard deviations of positions learned.

PR #95, opened 2020-10-29
Add transforms, features, and Claudius PR

KonstantinWilleke:
PR that has accumulated a bunch of small features.

- Scale transformation for the NeuroNormalizer in transforms.py
- solving a bug when normalizing the behavior states, and giving the eye_position the right name (pupil_center)
- adding a 5x5 laplace filter for the Laplace regularizer
- initializing a FileTree Dataset without behavior
- adding a new regularizer to the gaussian readout, that just returns the standard deviations of positions learned.

- Add the deterministic gaussian readout of Claudius


KonstantinWilleke:
This PR is a replacement for https://github.com/sinzlab/neuralpredictors/pull/89.
All issues that were mentioned there are addressed.

neuralpredictors/regularizers.py
```diff
@@ -17,6 +17,31 @@ def laplace():
     return np.array([[0, -1, 0], [-1, 4, -1], [0, -1, 0]]).astype(np.float32)[None, None, ...]
 
 
+def laplace5x5():
```
eywalker, line 20: Where did the definition of these filters come from? It actually appears to LoG, not Laplace filters.

neuralpredictors/regularizers.py
```diff
@@ -17,6 +17,31 @@ def laplace():
...
+    return np.array([[0, 0, 1, 0, 0],
+                     [0, 1, 2, 1, 0],
+                     [1, 2, -16, 2, 1],
+                     [0, 1, 2, 1, 0],
+                     [0, 0, 1, 0, 0]]).astype(np.float32)[None, None, ...]
+
+
+def laplace7x7():
+    """
+    Returns a 5x5 laplace filter.
```
eywalker, line 34: update this to say 7x7

neuralpredictors/data/transforms.py
```diff
@@ -298,24 +299,24 @@ def __init__(self, data, stats_source="all", exclude=None):
...
-            self._eye_std = np.array(
-                data.statistics["eye_position"][stats_source]["std"]
-            )
-            transforms["eye_position"] = lambda x: (x - self._eye_mean) / self._eye_std
-            itransforms["eye_position"] = lambda x: x * self._eye_std + self._eye_mean
 
+        # -- behavior
+        transforms['behavior'] = lambda x: x
+
+        if "pupil_center" in data.data_keys:
```
eywalker, line 309: make this work with either "pupil_center" or "eye_position" and not just overwrite. Static datasets use "pupil_center" and movie datasets are using "eye_position"

neuralpredictors/data/transforms.py
```diff
@@ -298,24 +299,24 @@ def __init__(self, data, stats_source="all", exclude=None):
...
+            self._eye_std = np.array(data.statistics["pupil_center"][stats_source]["std"])
+            transforms["pupil_center"] = lambda x: (x - self._eye_mean) / self._eye_std
+            itransforms["pupil_center"] = lambda x: x * self._eye_std + self._eye_mean
+
+        if "behavior" in data.data_keys:
             s = np.array(data.statistics["behavior"][stats_source]["std"])
 
             # TODO: same as above - consider other baselines
-            threshold = 0.01 * s.mean()
+            threshold = 0.01 * s
```
eywalker, line 319: this definition doesn't make sense as `s > 0.01 * s` is true all the time provided that `s` is a positive array (which it is as it's std values), so either completely get rid of this thresholding behavior, or come up with a meaningful thresholding scheme.

neuralpredictors/data/transforms.py
```diff
@@ -386,6 +389,8 @@ def __call__(self, x):
             "responses": self.transforms["responses"](key_vals["responses"]),
```
eywalker, line 389: why apply transforms given that `transforms` for `responses`, `behavior` and `pupile_center` are set as no-op. Do you expect these values to be overridden?

eywalker (changes requested):

eywalker (approved):

PR #105, opened 2021-01-21
Check for consistency in "add_neuron_meta"

kklurz:
The `add_neuron_meta` method in the FileTreeDataset does not check for consistency of the total number of neurons possible in the dataset and the actual number of neurons in the instantiation of the dataset. In the case of active subsampling transforms in the dataset instance, this can result in an error of inconsistent neuron indices later on which is very hard to debug.


neuralpredictors/data/datasets.py
```diff
@@ -706,17 +680,18 @@ def add_neuron_meta(self, name, animal_id, session, scan_idx, unit_id, values, f
             values:     new meta information. First dimension must refer to neurons.
             fill_missing:   fill the values of the new attribute with NaN if not provided
         """
+        with HiddenPrints():
+            temp = NewFileTreeDataset(self.dirname, "responses")
```
fabiansinz, line 684: I would do `NewFileTreeDataset(self.dirname, *self.data_keys)`.

eywalker: Per discussion, we suggested that a new context manager is implemented to temporally turn off the transforms

mohammadbashiri: Thanks for the suggestion. It's implemented.

neuralpredictors/data/datasets.py
```diff
@@ -706,28 +680,31 @@ def add_neuron_meta(self, name, animal_id, session, scan_idx, unit_id, values, f
...
-                self.neurons.scan_idx,
-                self.neurons.unit_ids,
+        with no_transforms(self):
+
+            if not self.n_neurons == len(values):
+                raise InconsistentDataException(
+                    f"Number of values ({len(values)}) and neurons in the datasets ({self.n_neurons}) is not consistent."
+                )
+
+            if not len(animal_id) == len(session) == len(scan_idx) == len(unit_id) == len(values):
```
fabiansinz, line 691: Unindent this!

kklurz: done

mohammadbashiri (approved):

PR #109, opened 2021-03-30
Add shifters

KonstantinWilleke:


ArneNx:
Please add some docstrings to this subpackage and the individual classes.

KonstantinWilleke:
> Please add some docstrings to this subpackage and the individual classes.

Ah, yes. Thanks!

neuralpredictors/layers/shifters/base.py
```diff
@@ -0,0 +1,8 @@
+class Shifter:
+    def __repr__(self):
+        s = super().__repr__()
+        s += " [{} regularizers: ".format(self.__class__.__name__)
+        ret = []
+        for attr in filter(lambda x: "gamma" in x, dir(self)):
+            ret.append("{} = {}".format(attr, getattr(self, attr)))
+        return s + "|".join(ret) + "]\n"
```
ArneNx, line 8: Would it make sense to make `Shifter` an abstract subclass of `nn.Module`? 
with additional `reuglarizer` and `initialize` methods that need to be implemented?

neuralpredictors/layers/shifters/static_affine.py
```diff
@@ -0,0 +1,45 @@
...
+import torch
+from torch import nn
+from torch.nn import ModuleDict
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class StaticAffine2dShifter(Shifter, ModuleDict):
```
eywalker, line 12: I'd prefer that we keep consistency in the order. In the `mlp.py` the `MLP` class came before `MLPShifter`. In the same spirit, I would prefer to see `StaticAffine2d` before `StaticAffine2dShifter` defined.

eywalker (changes requested):

neuralpredictors/layers/shifters/base.py
KonstantinWilleke: yes, sounds good. will do

neuralpredictors/layers/shifters/mlp.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class MLP(Shifter):
+    def __init__(self, input_features=2, hidden_channels=10, shift_layers=1, **kwargs):
+        """
+        Multi-layer perceptron shifter
+        Args:
+            input_features (int): number of input features, defaults to 2.
+            hidden_channels (int): number of hidden units.
+            shift_layers(int): number of hidden layers.
```
eywalker, line 19: I don't think this is an accurate description. Shift layers is literally the number of the layers including the hidden but also the mandatory output layer.

neuralpredictors/layers/shifters/mlp.py
```diff
@@ -0,0 +1,69 @@
...
+            **kwargs:
+        """
+        super().__init__()
+
```
eywalker, line 23: I recommend the following implementation to simplify:

```python
prev_output = input_features
feat = []
for _ in range(shift_layers - 1):
    feat.extend([nn.Linear(prev_output, hidden_channels), nn.Tanh()])
    prev_output = hidden_channels

feat.extend([nn.Linear(prev_output, 2), nn.Tanh()])
self.mlp = nn.Sequential(*feat)
```

eywalker: Alternatively, 

```python
prev_output = input_features
feat = []
out_features = [hidden_channels] * (shift_layers-1) + [2]  # yes this actually works
for output in out_features:
    feat.extend([nn.Linear(prev_output, output), nn.Tanh()])
    prev_output = output

self.mlp = nn.Sequential(*feat)
```

neuralpredictors/layers/shifters/static_affine.py
```diff
@@ -0,0 +1,59 @@
...
+import torch
+from torch import nn
+from torch.nn import ModuleDict
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class StaticAffine2d(Shifter):
```
eywalker, line 12: I don't think this would work - perhaps I mistakingly replaced the superclass here. I'm pretty sure this was meant to inherit from `nn.Linear`

eywalker (changes requested):

PR #116, opened 2021-03-30
Add relevant measures from other places

MaxFBurg:
Add measures from https://github.com/KonstantinWilleke/nnvision/blob/master/nnvision/utility/measures.py and https://github.com/sinzlab/nndichromacy/blob/435e521cab545b9ea4b594e9736956d8bb17947e/nndichromacy/utility/measures.py#L532


neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +14,157 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+
+    Args:
+        repeated_outputs (array-like): numpy array with shape (images, repeats, responses), or a list of lists of
+            responses (with len: num of repeats) per image.
+
+    Returns:
+        array: Oracle correlations per neuron
+    """
+
+    if len(repeated_outputs.shape) == 3:
```
eywalker, line 64: what is this checking?

neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +14,157 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+
+    Args:
+        repeated_outputs (array): numpy array with shape (images, repeats, responses), or a list of lists of
+            responses (with len: num of repeats) per image.
+
+    Returns:
+        array: Corrected oracle correlations per neuron
+    """
+
+    ImgVariance = []
```
eywalker, line 106: variable name convention is not Pythonic here - use `total_var` instead of `TotalVar`, `img_variance` instead of `ImgVariance` and so on.

neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +14,157 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+        targets (list): Neuronal responses (ground truth) to image repeats. Dimensions:
+            [num_images] np.array(num_repeats, num_neurons)
+        outputs (list): Model predictions to the repeated images, with an identical shape as the targets
+        return_exp_var (bool): returns the fraction of explainable variance per neuron if set to True
+    Returns:
+        FEVe (np.array): the fraction of explainable variance explained per neuron
+        --- optional: FEV (np.array): the fraction
+    """
+
+    ImgVariance = []
```
eywalker, line 130: same as above - would be great if you can use Pythonic naming conventions with `snake_case` instead of `CamelCase`

neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +14,157 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+    repeated_outputs = anscombe(repeated_outputs)
+    mu = np.array([np.mean(repeats, axis=0) for repeats in repeated_outputs])
+    mu_bar = np.mean(mu, axis=0)
+    sigma_2 = np.array([np.var(repeats, ddof=1, axis=0) for repeats in repeated_outputs])
+    sigma_2_bar = np.mean(sigma_2, axis=0)
+    snr = (1 / mu.shape[0] * np.sum((mu - mu_bar) ** 2, axis=0)) / sigma_2_bar
+    return snr if per_neuron else np.mean(snr)
+
+
+def anscombe(x):
```
eywalker, line 168: This is not the only place where Anscombe transformation occurs - let's decide on where to house this function.

eywalker (changes requested):

neuralpredictors/measures/np_functions.py
MaxFBurg: @KonstantinWilleke I thought this checks if it is a numpy array or a list, but not sure?

KonstantinWilleke: repeated_outputs can be an array of arrays (shape = (n_images)), or a single array of shape `(images, repeats, responses)`. That's what the check is for. Since the number of repeats is usually variable, the first type is more common.

neuralpredictors/measures/np_functions.py
MaxFBurg: Following Konstantin L.'s suggestion I'll move it to neuralpredictors.utils

neuralpredictors/measures/np_functions.py
mohammadbashiri: why not just use the code under `else`? that would work for both cases, right?

neuralpredictors/measures/np_functions.py
```diff
@@ -1,5 +1,8 @@
 import logging
 
+import numpy as np
+from neuralpredictors.utils import anscombe
```
eywalker, line 4: could you make this a relative import? Here it should be 

```python
from ..utils import anscombe
```

neuralpredictors/measures/np_functions.py
eywalker: Hm ok - then this check is rather crude for that but I'm ok with it. Could you add an explicit comment to state what this is checking?

mohammadbashiri (approved):

neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +15,152 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+        outputs (list): Model predictions to the repeated images, with an identical shape as the targets
+        return_exp_var (bool): returns the fraction of explainable variance per neuron if set to True
+    Returns:
+        FEVe (np.array): the fraction of explainable variance explained per neuron
+        --- optional: FEV (np.array): the fraction
+    """
+
+    img_var = []
+    pred_var = []
+    for i, _ in enumerate(targets):
```
eywalker, line 133: rather than doing `targets[i]` consider rewriting as follows:

```python
from itertools import count
for i, target, prediction in zip(count(), targets, predictions):
    pred_var.append((target - prediction) **2)
    img_var.append(np.var(targets, axis=0, ddof=1))
```

neuralpredictors/measures/np_functions.py
mohammadbashiri: why not just use the code under `else`? that would work for both cases, right?

eywalker (changes requested):

neuralpredictors/measures/np_functions.py
MaxFBurg: Thank you guys, I'll remove the code under if and just keep the code under else

neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +19,148 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+
+    Args:
+        repeated_outputs (array-like): numpy array with shape (images, repeats, responses), or a list of lists of
+            responses (with len: num of repeats) per image.
+
+    Returns:
+        array: Oracle correlations per neuron
+    """
+
+    if len(repeated_outputs.shape) == 3:
```
eywalker, line 65: Also this line would fail `repeated_outputs` is really a list of list so this should be changed. I suggest you do a check against the actual type of the object instead.

neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +19,148 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+        _, r, n = repeated_outputs.shape
+        oracles = (repeated_outputs.mean(axis=1, keepdims=True) - repeated_outputs / r) * r / (r - 1)
```
eywalker, line 67: Hm... with this, I'd really like for this duplication to be removed. Could you just handle the case of `len(repeated_outputs.shape)==3` as follows:

```
if len(repeated_outputs.shape) == 3:
    # repeated_outputs is not a stack of entries, so turn it into a single element list
    repeated_outputs = [repeated_outputs]

# Now just use the code found under else clause and the rest should work
```

neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +19,148 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+    return var_mean / np.sqrt(var_mean * (var_mean + var_noise))
+
+
+def oracle_corr(repeated_outputs: ArrayLike) -> np.ndarray:
+    """
+    Compute the oracle correlations per neuron.
+
+    Args:
+        repeated_outputs (array-like): numpy array with shape (images, repeats, responses), or a list of lists of
+            responses (with len: num of repeats) per image.
```
eywalker, line 59: Actually, it's not clear what's the expected structure for the list of lists. Is the first list supposed to be `images` or `repeats`?

eywalker (changes requested):

neuralpredictors/measures/np_functions.py
MaxFBurg: I'm not sure if that would work. I think repeated_outputs is a list containing for each image an array of shape (repeats, responses). So I guess it should work with just the code under the else clause: iterate over the list (i.e. for each image). Like this:
`oracles = []
    for outputs in repeated_outputs:
        r, n = outputs.shape
        # compute the mean over repeats, for each neuron
        mu = outputs.mean(axis=0, keepdims=True)
        # compute oracle predictor
        oracle = (mu - outputs / r) * r / (r - 1)

        if np.any(np.isnan(oracle)):
            logger.warning(
                "{}% NaNs when calculating the oracle. NaNs will be set to Zero.".format(
                    np.isnan(oracle).mean() * 100
                )
            )
            oracle[np.isnan(oracle)] = 0

        oracles.append(oracle)
    return corr(np.vstack(repeated_outputs), np.vstack(oracles), axis=0)`

MaxFBurg: Ah, I missed that one, I changed only one occurrence of that structure in l. 40 🤦🏻‍♂️ I'll push the updates to this code portion, too

neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +19,136 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+            logger.warning(
+                "{}% NaNs when calculating the oracle. NaNs will be set to Zero.".format(np.isnan(oracle).mean() * 100)
+            )
+            oracle[np.isnan(oracle)] = 0
+
+        oracles.append(oracle)
+    return corr(np.vstack(repeated_outputs), np.vstack(oracles), axis=0)
+
+
+def explainable_var(repeated_outputs: ArrayLike, eps: int = 1e-9) -> np.ndarray:
```
eywalker, line 83: what's the reason for supporting a list of arrays and just simply a full array? Someone could just go ahead and stack the list of arrays into a 3d array right?

neuralpredictors/measures/np_functions.py
```diff
@@ -12,10 +19,136 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+            image an array of shape (repeats, responses).
+
+    Returns:
+        array: Corrected oracle correlations per neuron
+    """
+
+    img_var = []
+    total_var = np.var(np.vstack(repeated_outputs), axis=0, ddof=1)
+    for out in repeated_outputs:
+        img_var.append(np.var(out, axis=0, ddof=1))
```
eywalker, line 98: Since you already perform `np.vstack(repeated_outputs)` in the line 96, you could have just performed `var(repeated_outputs, axis=0, ddof=1)` to directly obtain `img_var`. That should be significantly more efficient than an explicit for loop.

eywalker (changes requested):
Generally if you either just drop support for list of np.arrays or just start out by vstacking the list of arrays, you can greatly simplify the implementation, and as far as I can just avoid the use of an explicit for loop. That'd help to significantly make the implementation more efficient.

neuralpredictors/measures/np_functions.py
KonstantinWilleke: the lists are of unequal sizes most of the time. that's why they can be lists. the 3d shape is (images, repeats, neurons), and the number of repeats can vary greatly. One could just fill it up with NaNs, or just keep the list structure. Personally I just dont like NaNs so much, because they dont give a sense of the the actual content of the array.

MaxFBurg: Does that comment refer to line 80 or 83? :)

Regarding l. 80 and previous: In the end we vstack anyways. So I am wondering if we couldn't do that in the first place, then avoid the loop (I think this is what @eywalker proposes). Or is there something that I miss @KonstantinWilleke ?

neuralpredictors/measures/np_functions.py
MaxFBurg: I think the replacement is `img_var = var(repeated_outputs, axis=1, ddof=1)` since the loop removes dimension 0 from `out`, then we're performing `var` over the "new" dimension 0 which is dimension 1 in `repeated_outputs`, right?

neuralpredictors/measures/np_functions.py
MaxFBurg: 👍

KonstantinWilleke (dismissed):

neuralpredictors/measures/np_functions.py
```diff
@@ -14,15 +21,137 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
 
     Returns: correlation array
     """
+
     y1 = (y1 - y1.mean(axis=axis, keepdims=True)) / (y1.std(axis=axis, keepdims=True, ddof=0) + eps)
     y2 = (y2 - y2.mean(axis=axis, keepdims=True)) / (y2.std(axis=axis, keepdims=True, ddof=0) + eps)
     return (y1 * y2).mean(axis=axis, **kwargs)
 
 
+def oracle_corr_corrected(repeated_outputs: ArrayLike) -> np.ndarray:
```
KonstantinWilleke, line 34: We discussed that the name of the functions could be more descriptive.
If you agree, we'd call this function `conservative_oracle`

neuralpredictors/measures/np_functions.py
```diff
@@ -14,15 +21,137 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+    var_noise, var_mean = [], []
+    for output in repeated_outputs:
+        var_noise.append(output.var(axis=0))
+        var_mean.append(output.mean(axis=0))
+    var_noise = np.mean(np.array(var_noise), axis=0)
+    var_mean = np.var(np.array(var_mean), axis=0)
+    return var_mean / np.sqrt(var_mean * (var_mean + var_noise))
+
+
+def oracle_corr(repeated_outputs: ArrayLike) -> np.ndarray:
```
KonstantinWilleke, line 57: Continuing the comment above, this function might best be called `jackknife_oracle`.

KonstantinWilleke (changes requested):

neuralpredictors/measures/np_functions.py
MaxFBurg: Done!

neuralpredictors/measures/np_functions.py
MaxFBurg: Done!

KonstantinWilleke (approved):

eywalker (approved):

PR #118, opened 2021-03-30
Clean up utils

kklurz:


neuralpredictors/measures/np_functions.py
```diff
@@ -19,3 +21,24 @@ def corr(y1, y2, axis=-1, eps=1e-8, **kwargs):
...
+
+    Args:
+        x: 1 D array or list
+            Array of numbers from which to calculate the Gini coefficient.
+
+    Returns: float
+            Gini coefficient
+    """
+    x = np.asarray(x)
+    if any(i < 0 for i in x):
```
eywalker, line 39: shouldn't this just be `np.any(i < 0)`?

eywalker (changes requested):

neuralpredictors/measures/np_functions.py
kklurz: good catch, I changed it

neuralpredictors/measures/np_functions.py
```diff
@@ -36,9 +37,9 @@ def gini(x):
             Gini coefficient
     """
     x = np.asarray(x)
-    if any(i < 0 for i in x):
+    if any(x < 0):
```
eywalker, line 40: You have to use `np.any` here. If `x` has a shape with more than one dimension, `any` would fail. I don't think there is any reason to assume any shape/dim constraint on `x` here. However, that means you would want to offer a choice to specify along which axis to perform Gini coefficient computation. I suggest the following implementation:

```python
def gini(x, axis=None):
     """
        axis: axis along which to compute gini. If None, then the array is flattened first.
     """
    x = np.asarray(x)
    if axis is None:
        x = x.flatten()
        axis = -1
    if np.any(x < 0):
        warnings.warn("Input x contains negative values")
    sorted_x = np.sort(x, axis=axis)
    n = x.shape[axis]
    cumx = np.cumsum(sorted_x, dtype=float, axis=axis)
    return (n + 1 - 2 * np.sum(cumx, axis=axis) / cumx.take(-1, axis=axis)) / n

eywalker (changes requested):

PR #121, opened 2021-03-30
Core module redesign

KonstantinWilleke:


neuralpredictors/layers/cores/conv2d.py
```diff
@@ -542,3 +541,165 @@ def initialize(self):
...
+            hidden_channels:    Number of hidden channels (i.e feature maps) in each hidden layer
+            input_kern:     kernel size of the first layer (i.e. the input layer)
+            hidden_kern:    kernel size of each hidden layer's kernel
+            layers:         number of layers
+            gamma_input:    regularizer factor for the input weights (default: LaplaceL2, see neuralpredictors.regularizers)
+            skip:           Adds a skip connection
+            final_nonlinearity: Boolean, if true, appends an ELU layer after the last BatchNorm (if BN=True)
+            elu_xshift, elu_yshift: final_nonlinearity(x) = Elu(x - elu_xshift) + elu_yshift
+            bias:           Adds a bias layer.
+            momentum:       BN momentum
```
ArneNx, line 583: `BN`? Does this mean batch-norm? In that case, what does that mean?

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -542,3 +541,165 @@ def initialize(self):
...
+    ):
+        """
+        Args:
+            input_channels:     Integer, number of input channels as in
+            hidden_channels:    Number of hidden channels (i.e feature maps) in each hidden layer
+            input_kern:     kernel size of the first layer (i.e. the input layer)
+            hidden_kern:    kernel size of each hidden layer's kernel
+            layers:         number of layers
+            gamma_input:    regularizer factor for the input weights (default: LaplaceL2, see neuralpredictors.regularizers)
+            skip:           Adds a skip connection
```
ArneNx, line 579: naming it either `add_skip` or `skip_connection` may make it clearer

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -542,3 +541,165 @@ def initialize(self):
...
+            if batch_norm:
+                layer["norm"] = nn.BatchNorm2d(hidden_channels, momentum=momentum)
+
+            if (final_nonlinearity or l < self.layers - 1) and not linear:
+                layer["nonlin"] = nn.ELU(inplace=True)
+
+            if (self.layers - l) <= self.n_se_blocks:
+                layer["seg_ex_block"] = SqueezeExcitationBlock(in_ch=hidden_channels, reduction=se_reduction)
+
+            self.features.add_module("layer{}".format(l), nn.Sequential(layer))
```
ArneNx, line 684: Consider splitting these up into different methods, e.g. `add_first_layer` and `add_other_layers`

neuralpredictors/layers/cores/conv2d.py
ArneNx: Generally, if I read momentum, I'd expect it to be lr-related.

neuralpredictors/layers/cores/conv2d.py
ArneNx: Also, what is the type here, and what is the actual meaning?

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -542,3 +541,165 @@ def initialize(self):
...
+        assert not depth_separable or not attention_conv, "depth_separable and attention_conv should not both be true"
+
+        regularizer_config = (
+            dict(padding=laplace_padding, kernel=input_kern)
+            if input_regularizer == "GaussianLaplaceL2"
+            else dict(padding=laplace_padding)
+        )
+        self._input_weights_regularizer = regularizers.__dict__[input_regularizer](**regularizer_config)
+
+        self.layers = layers
```
ArneNx, line 618: I'd call it `self.num_layers` and `num_layers` or something. To make clear that this is an integer.

ArneNx (changes requested):

neuralpredictors/layers/cores/conv2d.py
KonstantinWilleke: It's an input to torch.nn.BatchNorm2d. I've clarified the doscstring.

neuralpredictors/layers/cores/conv2d.py
KonstantinWilleke: The problem with renaming the attributes is the backwards compatibility. model_configs that have skip should continue to work. But I agree that this is a very suboptimal name

KonstantinWilleke: The type is int, and the intention is to add skip connections to the n penultimate layers

ArneNx: I see the problem.

ArneNx (commented):
I would still split up the "first layer" and "other layers" construction into separate functions.
Apart from that, it looks good to me

KonstantinWilleke:
comments from the code review:
- reorder the init methods
- remove base class methods for layer init.

neuralpredictors/layers/squeeze_excitation.py
```diff
@@ -0,0 +1,34 @@
...
+class SqueezeExcitationBlock(nn.Module):
+    def __init__(self, in_ch, reduction=16):
+        """
+        A squeeze and excitation block as proposed by https://arxiv.org/abs/1709.01507
+        Args:
+            in_ch (int): number of input channels
+            reduction (int): reduction factor to calculate the output channels.
+        """
+        super().__init__()
+        self.se = nn.Sequential(
```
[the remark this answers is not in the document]
KonstantinWilleke: yes exactly

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -97,91 +102,124 @@ def __init__(
...
+        self.add_subsequent_layers()
+        self.initialize()
+
+        if depth_separable:
+            self.ConvLayer = DepthSeparableConv2d
+        elif attention_conv:
+            self.ConvLayer = self.AttentionConvWrapper
         else:
-            self.stack = [*range(self.layers)[stack:]] if isinstance(stack, int) else stack
+            self.ConvLayer = nn.Conv2d
```
cblessing24, line 150: `self.ConvLayer` is defined here but `self.add_subsequent_layers` is called beforehand and uses `self.ConvLayer` which doesn't exist yet as far as I can see. How does that work?

KonstantinWilleke: oh yes you are right. fixed!

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -51,6 +54,10 @@ def __init__(
         input_regularizer="LaplaceL2",
         stack=None,
         use_avg_reg=True,
+        depth_separable=False,
+        attention_conv=False,
+        linear=False,
+        stride=1,
```
cblessing24, line 60: Could you please document these arguments in the docstring?

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -97,91 +102,124 @@ def __init__(
...
-            self.stack = range(self.layers)
+            self.stack = range(self.num_layers)
+        else:
+            self.stack = [*range(self.num_layers)[stack:]] if isinstance(stack, int) else stack
+        self.linear = linear
+
+        if depth_separable:
+            self.ConvLayer = DepthSeparableConv2d
+        elif attention_conv:
+            self.ConvLayer = self.AttentionConvWrapper
```
cblessing24, line 143: Maybe add a warning if someone tries to pass `depth_separable` **and** `attention_conv`.

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -203,9 +241,9 @@ def group_sparsity(self):
         Sparsity regularization on the filters of all the conv2d layers except the first one.
         """
         ret = 0
-        for l in range(1, self.layers):
+        for l in range(1, self.num_layers):
             ret = ret + self.features[l].conv.weight.pow(2).sum(3, keepdim=True).sum(2, keepdim=True).sqrt().mean()
```
cblessing24, line 245: I would iterate over features directly.

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -97,91 +102,124 @@ def __init__(
...
-        if layers > 1 or final_nonlinearity:
-            layer["nonlin"] = AdaptiveELU(elu_xshift, elu_yshift)
+                layer["norm"] = nn.BatchNorm2d(
+                    self.hidden_channels, momentum=self.momentum, affine=self.bias and self.batch_norm_scale
+                )
+                if self.bias:
+                    if not self.batch_norm_scale:
+                        layer["bias"] = Bias2DLayer(self.hidden_channels)
+                elif self.batch_norm_scale:
+                    layer["scale"] = Scale2DLayer(self.hidden_channels)
```
cblessing24, line 172: This whole block is shared between this method and the `add_subsequent_layers` method. Therefore I propose putting it into a separate method. `SE2dCore` also uses this exact block of code if I am not mistaken.

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -292,122 +310,97 @@ def __init__(
...
-                elif batch_norm_scale:
-                    layer["scale"] = ScaleLayer(channels=hidden_channels)
-
-        if layers > 1 or final_nonlinearity:
-            layer["nonlin"] = AdaptiveELU(elu_xshift, elu_yshift)
+                if self.bias:
+                    if not self.batch_norm_scale:
+                        layer["bias"] = self.biaslayer_cls(channels=self.hidden_channels)
+                elif self.batch_norm_scale:
+                    layer["scale"] = self.scalelayer_cls(channels=self.hidden_channels)
```
cblessing24, line 359: This block of code is also shared similarly to the one in `Stacked2dCore`.

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -97,91 +102,124 @@ def __init__(
...
+            self.stack = [*range(self.num_layers)[stack:]] if isinstance(stack, int) else stack
+        self.linear = linear
+
+        if depth_separable:
+            self.ConvLayer = DepthSeparableConv2d
+        elif attention_conv:
+            self.ConvLayer = self.AttentionConvWrapper
         else:
-            self.stack = [*range(self.layers)[stack:]] if isinstance(stack, int) else stack
+            self.ConvLayer = nn.Conv2d
```
cblessing24, line 145: I would have preferred to have the `ConvLayer` as an argument to the `__init__` or a class attribute but I guess that is not possible without changing the interface of the class.

cblessing24 (changes requested):

neuralpredictors/layers/cores/conv2d.py
KonstantinWilleke: yes very good point. I've refactored it. Do you think it makes sense now?

cblessing24: Yes much better but I think we can do even better. I think you can use the new `add_activation` method in `RotationEquivariant2dCore` as well. 
Also the code that adds the batch norm layer in `RotationEquivariant2dCore` is almost the same as the one in `Stacked2dCore`. The only difference is that `RotationEquivariant2dCore` uses instance variables for the different layer classes instead of using the classes directly like `Stacked2dCore` does. Could we also use variables in the `add_bn_layer` method? Then we could use that method in `RotationEquivariant2dCore` as well.

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -542,3 +527,101 @@ def initialize(self):
...
+                                stack = -1 will only select the last layer as the readout layer.
+                                stack of -2 will read out from the last two layers.
+                                And stack of 1 will read out from layer 1 (0 indexed) until the last layer.
+            se_reduction:   Int, Reduction of channels for global pooling of the Squeeze and Excitation Block.
+            n_se_blocks:    Int, number of squeeze and excitation blocks. Inserted from the last layer
+                              Examples: layers=4, n_se_blocks=2:
+                                => layer0 -> layer1 -> layer2 -> SEblock -> layer3 -> SEblock
+            depth_separable: Boolean, if True, uses depth-separable convolutions in all layers after the first one.
+            attention_conv: Boolean, if True, uses self-attention instead of convolution for all layers after the first one.
+            linear:         Boolean, if True, removes all nonlinearities
```
kklurz, line 587: The docstring does not match the actual arguments of the function.

KonstantinWilleke: thanks, I updated it!

cblessing24 (approved):

kklurz (approved):

ArneNx (approved):

PR #122, opened 2021-03-30
Adds AdvancedTracker

ArneNx:
Adds a more advanced, universal tracker that offers many useful features (see documentation).

### DIFF
--- a/neuralpredictors/training/tracking.py
+++ b/neuralpredictors/training/tracking.py
@@ -1,9 +1,10 @@
 import copy
 import time
 import logging
-from collections import defaultdict
-from typing import Dict, Tuple, List
+from collections import defaultdict, abc
+from typing import Dict, Tuple, List, Mapping, Union, Optional, Sequence
 import numpy as np
+from tqdm import tqdm
 
 from .utils import deep_update
 
@@ -244,16 +245,18 @@ class AdvancedTracker(Tracker):
     ```
     """
 
-    def __init__(self, main_objective: Tuple[str, ...] = (), **objectives):
+    def __init__(self, main_objective: Tuple[str, ...] = (), **objectives: Mapping) -> None:
         """
         In principle, `objectives` is expected to be a dictionary of dictionaries
-        The hierarchy can in principle be arbitrary deep.
+        The hierarchy can in principle be arbitrarily deep.
         The only restriction is that the lowest level has to be a dictionary with values being
         either a numerical value which will be interpreted as the initial value for this objective
         (to be accumulated manually) or a callable (e.g. a function) that returns the objective value.
 
         Args:
             main_objective: key of the main objective that is used to e.g. decide lr reductions
+                            can be something like `("img_classification","accuracy")` to always look at accruacy.
+                            Can be combined with a setting specific key in `get_current_main_objective(...)`
             **objectives: e.g. {"dataset": {"objective1": o_fct1, "objective2": 0, "normalization": 0},...}
                            or {"dataset": {"task_key": {"objective1": o_fct1, "objective2": 0},...},...}
         """
@@ -261,10 +264,10 @@ class AdvancedTracker(Tracker):
         self.log = self._initialize_log(objectives)
         self.time = []
         self.main_objective = main_objective
-        self.epoch = -1
+        self.epoch = -1  # will be increased to 0 in next line
         self.start_epoch()
 
-    def add_objectives(self, objectives: Dict, init_epoch: bool = False):
+    def add_objectives(self, objectives: Mapping, init_epoch: bool = False) -> None:
         """
         Add new objectives (with initial values) to the logs.
         Args:
@@ -277,14 +280,14 @@ class AdvancedTracker(Tracker):
             self._initialize_epoch(new_log, objectives)
         deep_update(self.log, new_log)
 
-    def start_epoch(self):
+    def start_epoch(self) -> None:
         """ Start a new epoch. Initialize each accumulation with its default value. """
         t = time.time()
         self.time.append(t)
         self.epoch += 1
         self._initialize_epoch(self.log, self.objectives)
 
-    def log_objective(self, value: float, key: Tuple[str, ...] = ()):
+    def log_objective(self, value: float, key: Tuple[str, ...] = ()) -> None:
         """
         Add a new entry to the logs
         Args:
@@ -296,7 +299,7 @@ class AdvancedTracker(Tracker):
         else:
             self._log_objective_callables(self.log, self.objectives)
 
-    def display_log(self, key=(), tqdm_iterator=None):
+    def display_log(self, key: Tuple[str, ...] = (), tqdm_iterator: Optional[tqdm] = None) -> None:
         """
         Display the current objective value of everything under `key`
         Args:
@@ -315,9 +318,9 @@ class AdvancedTracker(Tracker):
         else:
             logger.info(current_log)
 
-    def check_isfinite(self, log: dict = None) -> bool:
+    def check_isfinite(self, log: Optional[Union[Mapping, Sequence]] = None) -> bool:
         """
-        Checks if all entries in `log` or `self.log` are finite.
+        Checks if all entries in `log` or (normalized) `self.log` are finite.
         Args:
             log: dict that is recursively searched for infinite entries
 
@@ -325,7 +328,7 @@ class AdvancedTracker(Tracker):
         """
         if log is None:
             log = self._normalize_log(self.log)
-        if isinstance(log, dict):
+        if isinstance(log, abc.Mapping):
             for k, l in log.items():
                 if not self._check_isfinite(l):
                     return False
@@ -333,8 +336,9 @@ class AdvancedTracker(Tracker):
             return np.isfinite(log).any()
         return True
 
-    def get_objective(self, log=None, key: Tuple[str, ...] = ()) -> np.array:
+    def get_objective(self, log: Optional[Union[Mapping, Sequence]] = None, key: Tuple[str, ...] = ()) -> np.array:
         """
+        Get the value of the objective that corresponds to the key.
         Args:
             log: log to retrieve objective from
             key: key to match with the log
@@ -349,14 +353,14 @@ class AdvancedTracker(Tracker):
         else:
             return log[key[0]]
 
-    def get_current_objective(self, key) -> float:
+    def get_current_objective(self, key: Tuple[str, ...]) -> float:
         return self.get_objective(self._normalize_log(self.log), key)[-1]
 
-    def get_current_main_objective(self, key) -> float:
+    def get_current_main_objective(self, key: Tuple[str, ...]) -> float:
         combined_key = key + self.main_objective if isinstance(key, tuple) else (key,) + self.main_objective
         return self.get_current_objective(combined_key)
 
-    def finalize(self):
+    def finalize(self) -> None:
         """ After training, normalize the log and save the total time. """
         self.time = np.array(self.time)
         self.time -= self.time[0]
@@ -372,7 +376,7 @@ class AdvancedTracker(Tracker):
         output = copy.deepcopy(self.__dict__)
         return output
 
-    def load_state_dict(self, tracker_dict: Dict):
+    def load_state_dict(self, tracker_dict: Dict) -> None:
         """
         Loads given state_dict from another tracker.
         Args:
@@ -402,16 +406,16 @@ class AdvancedTracker(Tracker):
         tracker.epoch = tracker_dict["epoch"]
         return tracker
 
-    def _initialize_log(self, objectives) -> Dict:
+    def _initialize_log(self, objectives: Mapping[str, any]) -> Dict:
         log = {}
         for key, objective in objectives.items():
-            if isinstance(objective, dict):
+            if isinstance(objective, abc.Mapping):
                 log[key] = self._initialize_log(objective)
             elif not callable(objective):
                 log[key] = []
         return log
 
-    def _initialize_epoch(self, log, objectives):
+    def _initialize_epoch(self, log: Union[Mapping, Sequence], objectives: Mapping[str, any]) -> None:
         """
         For each key in `objectives`, go through log and append a new entry to its list.
         The entry reflects the default value saved in `objectives`.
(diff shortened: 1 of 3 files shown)


neuralpredictors/training/utils.py
```diff
@@ -0,0 +1,20 @@
+import collections
+from typing import Dict
+
+
+def deep_update(d: Dict, u: Dict) -> Dict:
```
cblessing24, line 5: I'd say `Mapping` would be the correct type for all parameters and the return value here because your code will work if `d` and `u` are of type `Mapping` and there is no guarantee that the return value will be of type `Dict`.

cblessing24: Or even more correct would be:
```
import collections
from typing import Dict, TypeVar

T = TypeVar("T", bound=collections.abc.Mapping)

def deep_update(d: T, u: collections.abc.Mapping) -> T:
    ...
```
This would tell a static type checker that the return value has to be of the same type as `d` and that `d` has to be a subclass of `Mapping`.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+                norm = norm[nonzero_start:]
+                for key, l in n_log.items():
+                    l = l[nonzero_start:]
+                    if isinstance(l, np.ndarray):
+                        n_log[key] = l / np.where(norm > 0, norm, np.ones_like(norm))
+            return n_log
+        else:
+            return np.array(log)
+
+    def _gather_log(self, log: Dict, key: Tuple[str, ...], index: int = -1) -> Dict:
```
cblessing24, line 489: Judging from the code the type of `log` can be `List` (or more generally `Sequence`) as well as `Dict` (or `Mapping`) but the signature only specifies `Dict`.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+        Disclaimer: this is not very well tested!
+        Args:
+            log:
+            objectives:
+        """
+        for key, objective in objectives.items():
+            if isinstance(objective, dict):
+                self._log_objective_callables(log[key], objective)
+            elif callable(objective):
+                log[key].append(objective())
```
cblessing24, line 455: I assume a callable objective is supposed to return the objective value, correct? Why is this value appended here instead of being added?

I would have expected something like `log[key][self.epoch] += objective()` here.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+        # normalize (if applicable) and turn into np.arrays:
+        n_log = self._normalize_log(self.log)
+        # flatten a subset of the dictionary:
+        current_log = self._gather_log(n_log, key, index=-1)
+        if tqdm_iterator:
+            tqdm_iterator.set_postfix(**current_log)
+        else:
+            logger.info(current_log)
+
+    def check_isfinite(self, log: dict = None) -> bool:
```
cblessing24, line 318: I think the type annotation for `log` should be `Optional[Dict]`.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+        """
+        Checks if all entries in `log` or `self.log` are finite.
```
cblessing24, line 320: Might be worth mentioning that the log is normalized if `self.log` is used.

neuralpredictors/training/tracking.py
cblessing24: Just realized that `log` can also be a Sequence. So the type annotation should be something like `Optional[Union[Sequence, Mapping]]`.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+        """
+        if log is None:
+            log = self._normalize_log(self.log)
+        if isinstance(log, dict):
+            for k, l in log.items():
+                if not self._check_isfinite(l):
+                    return False
+        else:
+            return np.isfinite(log).any()
+        return True
```
cblessing24, line 334: I think this return statement is unreachable.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+
+    def get_objective(self, log=None, key: Tuple[str, ...] = ()) -> np.array:
+        """
```
cblessing24, line 337: A one sentence summary in the docstring would be nice.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+        if log is None:
+            log = self._normalize_log(self.log)
+        if len(key) > 1:
+            return self.get_objective(log[key[0]], key[1:])
+        else:
+            return log[key[0]]
+
+    def get_current_objective(self, key) -> float:
+        return self.get_objective(self._normalize_log(self.log), key)[-1]
```
cblessing24, line 353: This will throw a `KeyError` if `key` doesn't specify enough levels which might be confusing.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+                "img_classification": {
+                    "loss": 0,
+                    "accuracy": 0,
+                    "normalization": 0,
+                },
+                "patience": 0,
+            },
+        }
+        tracker = AdvancedTracker(
+            main_objective=("img_classification", "accuracy"), **objectives
```
cblessing24, line 215: Why is `main_objective` not starting with the `"training"` key?

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+            log = self._normalize_log(self.log)
+        if len(key) > 1:
+            return self.get_objective(log[key[0]], key[1:])
+        else:
+            return log[key[0]]
+
+    def get_current_objective(self, key) -> float:
+        return self.get_objective(self._normalize_log(self.log), key)[-1]
+
+    def get_current_main_objective(self, key) -> float:
```
cblessing24, line 355: A bit confused as to why we need `key` here. Doesn't `self.main_objective` correspond to a single sequence of values in the log?

cblessing24 (changes requested):
I quite like this implementation. It is very flexible and makes nice use of recursion. :+1: 

Some general comments:
* I would either annotate everything (also `None` if it returns nothing) in a method signature or nothing.
* The code could be made more flexible by checking for a broader type that still works (i.e. `Dict` vs. `Mapping`).
* Some public methods are missing docstrings.
* Wording in docstrings is inconsistent (imperative vs. non-imperative).

I also left some minor comments in the code.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +153,360 @@ def asdict(self, time_key="time", make_copy=True):
...
+            log:
+            objectives:
+        """
+        for key, objective in objectives.items():
+            if isinstance(objective, dict):
+                self._log_objective_callables(log[key], objective)
+            elif callable(objective):
+                log[key].append(objective())
+
+    def _normalize_log(self, log: [Dict, List]) -> [Dict, np.array]:
```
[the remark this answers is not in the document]
ArneNx: That's actually what I meant. Thanks for pointing that out.

neuralpredictors/training/tracking.py
ArneNx: I added a comment about that in the docstring of the init now.

neuralpredictors/training/tracking.py
ArneNx: See above (comment in the init).

neuralpredictors/training/tracking.py
ArneNx: What would you suggest should be thrown instead?

neuralpredictors/training/tracking.py
cblessing24: Ah, yes, you are right. My bad.

neuralpredictors/training/tracking.py
cblessing24: I'd say it should raise `KeyError` if `key` is not in the log and a `ValueError` with a descriptive error message if `key` is in the log but doesn't specify a item in the log on the lowest level.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +154,369 @@ def asdict(self, time_key="time", make_copy=True):
...
+        """
+        In principle, `objectives` is expected to be a dictionary of dictionaries
+        The hierarchy can in principle be arbitrarily deep.
+        The only restriction is that the lowest level has to be a dictionary with values being
+        either a numerical value which will be interpreted as the initial value for this objective
+        (to be accumulated manually) or a callable (e.g. a function) that returns the objective value.
+
+        Args:
+            main_objective: key of the main objective that is used to e.g. decide lr reductions
+                            can be something like `("img_classification","accuracy")` to always look at accruacy.
```
cblessing24, line 258: There is a typo in accuracy.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +154,369 @@ def asdict(self, time_key="time", make_copy=True):
...
+        """
+        if norm is not None:
+            nonzero_start = (norm != 0).argmax(axis=0)
+            norm = norm[nonzero_start:]
+            for key, l in n_log.items():
+                l = l[nonzero_start:]
+                if isinstance(l, np.ndarray):
+                    n_log[key] = l / np.where(norm > 0, norm, np.ones_like(norm))
+
+    def _gather_log(self, log: Union[Mapping, Sequence], key: Tuple[str, ...], index: int = -1) -> Mapping[str, str]:
```
cblessing24, line 499: I would say `Dict[str, str]` as the type of the return value because the method always returns a dictionary. That way the caller doesn't have to check whether it is a `Dict` or not (if that should matter to the caller).

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +154,369 @@ def asdict(self, time_key="time", make_copy=True):
...
+
+        """
+        self.main_objective = tracker_dict["main_objective"]
+        self.objectives = tracker_dict["objectives"]
+        self.log = tracker_dict["log"]
+        self.time = tracker_dict["time"]
+        self.epoch = tracker_dict["epoch"]
+
+    @classmethod
+    def from_dict(cls, tracker_dict: Dict) -> "AdvancedTracker":
```
cblessing24, line 393: You can put `from __future__ import annotations` at the top of the file and then you don't need the quotes around the return type.

neuralpredictors/training/tracking.py
```diff
@@ -1,6 +1,15 @@
-from collections import defaultdict
-import numpy as np
+import copy
 import time
+import logging
+from collections import defaultdict, abc
+from typing import Dict, Tuple, List, Mapping, Union, Optional, Sequence
+import numpy as np
+from tqdm import tqdm
+from __future__ import annotations
```
cblessing24, line 8: This needs to be the first line of code in the file (i.e. right after the docstring). That is why the tests are failing.

neuralpredictors/training/tracking.py
```diff
@@ -1,6 +1,15 @@
-from collections import defaultdict
```
cblessing24, line 1: Could you please add a docstring for the module?

cblessing24 (changes requested):

eywalker (changes requested):

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +155,372 @@ def asdict(self, time_key="time", make_copy=True):
...
+        # normalize (if applicable) and turn into np.arrays:
+        n_log = self._normalize_log(self.log)
+        # flatten a subset of the dictionary:
+        current_log = self._gather_log(n_log, key, index=-1)
+        if tqdm_iterator:
+            tqdm_iterator.set_postfix(**current_log)
+        else:
+            logger.info(current_log)
+
+    def check_isfinite(self, log: Optional[Union[Mapping, Sequence]] = None) -> bool:
```
KonstantinWilleke, line 322: not sure if I understand this correctly - but this shouldn't this rather be a private method?

KonstantinWilleke: Also, for readability, I'd move this method down to the other class methods, then the ordering of methods becomes more intuitive.

neuralpredictors/training/tracking.py
```diff
@@ -146,3 +155,372 @@ def asdict(self, time_key="time", make_copy=True):
...
+        The hierarchy can in principle be arbitrarily deep.
+        The only restriction is that the lowest level has to be a dictionary with values being
+        either a numerical value which will be interpreted as the initial value for this objective
+        (to be accumulated manually) or a callable (e.g. a function) that returns the objective value.
+
+        Args:
+            main_objective: key of the main objective that is used to e.g. decide lr reductions
+                            can be something like `("img_classification","accuracy")` to always look at accuracy.
+                            Can be combined with a setting specific key in `get_current_main_objective(...)`
+            **objectives: e.g. {"dataset": {"objective1": o_fct1, "objective2": 0, "normalization": 0},...}
```
KonstantinWilleke, line 261: Could you extend the docstring a bit to elaborate on the `normalization` I have troubles understanding what it does and how it is used.

KonstantinWilleke (changes requested):

neuralpredictors/training/tracking.py
ArneNx: This is actually used from the outside in some cases, so I'd keep it public. 
I moved it down though.

neuralpredictors/training/tracking.py
ArneNx: good point

cblessing24 (approved):

KonstantinWilleke (approved):

eywalker (approved):

cblessing24 (approved):

PR #123, opened 2021-04-01
New MultiReadout structure

kklurz:


kklurz:
So this is a draft of how the new MultiReadout structure could look like. It's not finished yet, tests are not expected to run etc. I just want to discuss with everyone the general idea so that I don't waste time implementing something thorouhly which might get discarded anyways

kklurz:
TODOs:
- I could not implement the regularization method for the Gaussian3D -> there is no featurel1 or any other method 
- No idea what the "RemappedGaussian2d" does. Whoever implemented this, needs to take over...
- The factorized readout has a different default for "average" in the regularizer than all other readouts. -> Do we want to break backwards compatibility or break consistency between readouts? 
- I'm not sure what the Pyramid readout does. There is basically none of the important methods implemented... On the other hand, the PointPyramid2d readout does not inherit from it, so it can not be meant as a base class for this?

neuralpredictors/layers/readouts/base.py
```diff
@@ -23,106 +23,34 @@ def initialize(self, *args, **kwargs):
...
-
-    @property
-    def normalized_spatial(self):
-        if self.normalize:
-            norm = self.spatial.pow(2).sum(dim=1, keepdim=True)
-            norm = norm.sum(dim=2, keepdim=True).sqrt().expand_as(self.spatial) + 1e-6
-            weight = self.spatial / norm
+        return super().__repr__() + " [{}]\n".format(self.__class__.__name__)
+
+    def reduction_method(self, reduction="mean", average=None):
```
eywalker, line 28: It tends to be helpful for method name to describe an action. Perhaps call this something like `resolve_reduction_method`.

neuralpredictors/layers/readouts/base.py
```diff
@@ -23,106 +23,34 @@ def initialize(self, *args, **kwargs):
...
+                f"Reduction method '{reduction}' is not recognized. Valid values are ['mean', 'sum', None]"
+            )
 
-    def initialize(self):
-        self.spatial.data.normal_(0, self.init_noise)
-        self.features.data.normal_(0, self.init_noise)
-        if self.bias is not None:
+    def initialize_bias(self, mean_activity=None):
+        if mean_activity is None:
+            warnings.warn("Readout is NOT initialized with mean activity!")
```
eywalker, line 50: Could you indicate that it is initialized to 0 instead?

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -1,107 +1,125 @@
...
+        self,
+        in_shape_dict,
+        n_neurons_dict,
+        base_readout=None,
+        mean_activity_dict=None,
+        clone_readout=False,
+        gamma_readout=1.0,
+        **kwargs
+    ):
         if self._base_readout is None:
```
eywalker, line 25: Could you add a comment here saying something like:
```
# The `base_readout` can be overridden only if the static property `_base_readout` is not set

File: neuralpredictors/layers/readouts/multi_readout.py [hunk 4]
```diff
@@ -1,107 +1,125 @@
...
+from .point_pooled import PointPooled2d
+from .gaussian import FullGaussian2d
+from .factorized import SpatialXFeatureLinear
+
+#################################################################
+#### MultiReadout Base Classes
+#################################################################
+
+
+class MultiReadoutBase(torch.nn.ModuleDict):
```
eywalker, line 12: Let's add docstring to describe the two base classes.

eywalker (changes requested):

ArneNx:
Please some documentation to the base classes

neuralpredictors/layers/readouts/gaussian.py
```diff
@@ -91,9 +94,8 @@ def initialize(self):
                 "fixed_sigma to True"
             )
         self.features.data.fill_(1 / self.in_shape[0])
-
         if self.bias is not None:
-            self.bias.data.fill_(0)
+            self.initialize_bias(mean_activity=mean_activity)
```
mohammadbashiri, line 98: What happens if `mean_activity` is passed but `bias=None`?

neuralpredictors/layers/readouts/base.py
```diff
@@ -19,110 +12,66 @@ class ConfigurationError(Exception):
 
 
 class Readout(nn.Module):
```
mohammadbashiri, line 14: It seems all the readouts have the `regularizer()` method. And I think the Multi-readout expects it. In that case, I suggest we add it to the base class where it returns `0.0`.

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -1,107 +1,147 @@
...
 
-    def __init__(self, in_shape, loaders, gamma_readout, clone_readout=False, **kwargs):
+    def __init__(
+        self,
+        in_shape_dict,
+        n_neurons_dict,
+        base_readout=None,
+        mean_activity_dict=None,
+        clone_readout=False,
+        gamma_readout=1.0,
```
mohammadbashiri, line 37: Several point here:

1. Why is there the assumption that `gamma_readout` applies to any readout?
2. why the inconsistency between `gamma_readout` and `reg_weight`? I guess due to backwards compatibility?

Suggestion: Let's leave any regulariziation-related argument in the kwargs. It can easily happen that a readout has multiple regularization each with their own specific regularization coefficient. Putting this argument in the kwargs generalizes the input arguments and also preserves backwards compatibility.

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -1,107 +1,147 @@
...
             if i == 0 or clone_readout is False:
                 self.add_module(
-                    k,
-                    self._base_readout(in_shape=in_shape, outdims=n_neurons, **kwargs),
+                    data_key,
+                    self._base_readout(
+                        in_shape=in_shape_dict[data_key],
+                        outdims=n_neurons_dict[data_key],
+                        mean_activity=mean_activity,
+                        reg_weight=gamma_readout,
```
mohammadbashiri, line 62: Based on the comment above this would be removed.

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -1,107 +1,147 @@
...
-        in_shape (list): shape of the input feature map [channels, width, height]
-        loaders (list):  a list of dataset objects
-        gamma_readout (float): regularisation term for the readout which is usally set to 0.0 for gaussian3d readout
-                               as it contains one dimensional weight
-
-    """
+    def initialize(self, mean_activity_dict=None):
+        for data_key, readout in self.items():
+            mean_activity = mean_activity_dict[data_key] if mean_activity_dict is not None else None
+            readout.initialize(mean_activity)
```
mohammadbashiri, line 83: Is there a reason we have these both as a method and also as input argument while initializing readout modules?

mohammadbashiri (changes requested):

neuralpredictors/layers/readouts/gaussian.py
kklurz: Every individual readout has this line in the initialize method, so it's taken care of:
if self.bias is not None:
    self.initialize_bias(mean_activity=mean_activity)

neuralpredictors/layers/readouts/base.py
kklurz: Good point. I would add a NotImplementedError version of the regularizer method though, instead of returning 0.0. What do you think? People can still easily just make it return 0 but they should be aware of it.

neuralpredictors/layers/readouts/multi_readout.py
kklurz: The reason behind the new name is just that gamma_raedout is a terrible name and I feel like now is the only time to change it. I like your comment about leaving regularization to the kwargs, what do other people think about it?

neuralpredictors/layers/readouts/multi_readout.py
kklurz: you mean line 62?

neuralpredictors/layers/readouts/multi_readout.py
kklurz: What do you mean with "these"?

neuralpredictors/layers/readouts/multi_readout.py
mohammadbashiri: yes exactly

neuralpredictors/layers/readouts/multi_readout.py
mohammadbashiri: when initializing the modules you have:

```python
self.add_module(
    data_key,
    self._base_readout(
        in_shape=in_shape_dict[data_key],
        outdims=n_neurons_dict[data_key],
        mean_activity=mean_activity,
        reg_weight=gamma_readout,
        **readout_kwargs,
        ),
    )
```

which has the `mean_activity=mean_activity` in the arguments. And every readout also has `self.initialize(mean_activity)` in their `__init__` method. Which means the `initialize()` method of `MultiReadoutBase` is redundant. right?

neuralpredictors/layers/readouts/gaussian.py
mohammadbashiri: But if someone passes the `mean_activity` without setting `bias=True` then it is gonna be ignored, right? I mean, in this scenario even though the user has passed the `mean_activity` the bias would be initialized randomly (or whatever the default is for Pytorch)
Also, now that I am thinking about this, what are the expected values of `bias`? is it supposed to be a boolean?

neuralpredictors/layers/readouts/base.py
mohammadbashiri: @eywalker @KonstantinWilleke what do you guys think?

neuralpredictors/layers/readouts/multi_readout.py
mohammadbashiri: @eywalker @KonstantinWilleke what do you guys think?

neuralpredictors/layers/readouts/base.py
mohammadbashiri: I agree better make it fail than silently potentially not work.

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -1,107 +1,147 @@
...
-        for k in self:
-            self[k].positive = value
+                self.add_module(data_key, ClonedReadout(self[original_readout]))
 
+        self.initialize(mean_activity_dict)
 
-class MultiplePointPyramid2d(MultiReadout):
-    _base_readout = PointPyramid2d
+    def prepare_readout_kwargs(self, i, data_key, first_data_key, **kwargs):
+        return kwargs
```
mohammadbashiri, line 73: This method seems redundant. Given the current functionality, is there a reason it's there? 
On the other hand, I feel like it should do more: get the `kwargs`, compare them to the arguments of the `_base_readout`, and then pass the relevant arguments with a warning, or raise an error.

neuralpredictors/layers/readouts/multi_readout.py
mohammadbashiri: I also have another question here. When initializing each readout module, you pass in `in_shape` and `outdims` followed by `mean_activity`. But then, for instance, in [FullGaussian2d](https://github.com/sinzlab/neuralpredictors/blob/4ac2ed8d6404eef52bdf619ed1a9fe565994f497/neuralpredictors/layers/readouts/gaussian.py#L205), the first three arguments are positional (i.e. must be passed) and one of them is `bias`. I know you said you tested the initialization and it works, but I am just wondering how? I thought initializing a FullGaussian2d should throw an error because `bias`  was not passed as input.

neuralpredictors/layers/readouts/gaussian.py
kklurz: If bias=False, there will be no initialization with 0 because there will be no bias. The argument `bias` to the `__init__` is boolean and says whether there is a bias or not. `self.bias` in the initialize, however, is either `None` (if `bias` was `False`) or a `torch.tensor` (if `bias` was `True`). `if self.bias is not None` (i.e. if `bias` was `True`), the so existing bias will be initialized with the mean activity (which defaults to None and then results in initialization with 0)

neuralpredictors/layers/readouts/multi_readout.py
kklurz: You are right, it is redundant for the first initialization after instantiation (which is 99% of the cases). I would still like to argue for implementing this method also on the multireadout level, such that one can use it to re-initialize all readouts by simply calling `model.readout.initialize()`. A good point to discuss is though, whether we want to turn the `mean_activity` of the readout's `__init__` into a class attribute and then use `self.mean_activity` in the `initialize` if it's own argument `mean_activity` is None

neuralpredictors/layers/readouts/multi_readout.py
kklurz: It's just there so that it can be defined in a more meaningful way in `MultiReadoutSharedParametersBase`

neuralpredictors/layers/readouts/multi_readout.py
kklurz: All arguments are passed as keyword arguments (`inshape`, `outdims`, `mean_activity`, `reg_weight`) and the `**readout_kwargs` results in keyword arguments as well. So if everything that is passed is a keyword argument, the order doesn't matter. If you choose a `base_readout` that needs a specific argument (like `bias` in `FullGaussian2d`) and you don't give it in the `readout_kwargs`, then of course it will give you an error

neuralpredictors/layers/readouts/base.py
```diff
@@ -19,110 +12,69 @@ class ConfigurationError(Exception):
...
-        return (
-            ("normalized " if self.normalize else "")
-            + self.__class__.__name__
-            + " ("
-            + "{} x {} x {}".format(*self.in_shape)
-            + " -> "
-            + str(self.outdims)
-            + ")"
-        )
+        return super().__repr__() + " [{}]\n".format(self.__class__.__name__)
```
KonstantinWilleke, line 27: A very minor point here. For readability, could you rearrange the methods? that would make understanding this base class a bit easier. I'd suggest:
- init
- regularizer
- apply_reduction
- resolve_reduction..
- __repr__

neuralpredictors/layers/readouts/multi_readout.py
KonstantinWilleke: Nice, I like the idea a lot andI agree with leaving it in the `kwargs`. And it is a good name to change the name to be more descriptive. I'd advocate that we rename `gamma_readout` into `gamma_features`, since this is what is usually being regularized in all of our readouts. (Except Factorized, but we can tackle this later on).

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -1,107 +1,147 @@
...
+import torch
+from .base import *
+from .point_pooled import *
+from .gaussian import *
+from .factorized import *
+from .attention import *
+from .pyramid import *
+
+
+class MultiReadoutBase(torch.nn.ModuleDict):
```
KonstantinWilleke, line 10: I'd find it more intuitive if this base class is also in `base.py`. And then I think we can get rid of `multi_readout.py`, and instead have everything in the `__init__.py` So the classes itself, like FullGaussian2d, as well as the instantiated MultiReadouts. Then everything can be conventiently imported from layers/readouts import *.

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -1,107 +1,147 @@
...
-    """
-    Instantiates multiple instances of FullGaussian2d Readouts
-    usually used when dealing with more than one dataset sharing the same core.
-
-    Args:
-        in_shape (list): shape of the input feature map [channels, width, height]
-        loaders (list):  a list of dataloaders
-        gamma_readout (float): regularizer for the readout
-    """
+class MultipleSpatialXFeatureLinear(MultiReadoutBase):
```
KonstantinWilleke, line 138: Could you also add a `MultipleFactorized2d` here? (or in the __init__, see comment above), then we have the important readouts already ready to import an use.

KonstantinWilleke (changes requested):

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -1,107 +1,147 @@
...
+    """
+
     _base_readout = None
 
-    def __init__(self, in_shape, loaders, gamma_readout, clone_readout=False, **kwargs):
+    def __init__(
+        self,
+        in_shape_dict,
+        n_neurons_dict,
+        base_readout=None,
```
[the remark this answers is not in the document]
kklurz: That was one of Edgar's comments. He wanted it as an class attribute for backwards compatibility but he wanted to also be able to simply pass it to the init so he doesn't need to define a new class for it

neuralpredictors/layers/readouts/multi_readout.py
kklurz: I don't think this is possible, it would be a circular import.

neuralpredictors/layers/readouts/__init__.py
```diff
@@ -0,0 +1,28 @@
+from .point_pooled import *
+from .gaussian import *
+from .factorized import *
+from .attention import *
+from .pyramid import *
+from .multi_readout import *
```
cblessing24, line 6: I would suggest using the actual names of the things you want to import instead of using `*`. This way of doing things might lead to hard to detect bugs in the future. For example it would be possible to import the same name from different modules without knowing.

neuralpredictors/layers/readouts/attention.py
```diff
@@ -7,11 +14,16 @@ def __init__(
...
         attention_layers=1,
+        mean_activity=None,
+        feature_reg_weight=1.0,
+        gamma_readout=None,  # depricated, use feature_reg_weight instead
         **kwargs,
     ):
         super().__init__()
         self.in_shape = in_shape
         self.outdims = outdims
+        self.feature_reg_weight = self.resolve_depricated_gamma_readout(feature_reg_weight, gamma_readout)
```
cblessing24, line 25: Should say "deprecated" instead of "depricated".

ArneNx:
Base class documentation looks good to me now.

KonstantinWilleke (approved):

mohammadbashiri (dismissed):

eywalker (approved):

PR #124, opened 2021-04-13
Eval state fix

KonstantinWilleke:
This PR fixes a bug. The user can specify a cuda device in two ways, either by `device = "cuda:0"`, which is recommended practice. But the user can also be lazy by running `device = "cuda"`. This will by default use the 0th-device.
But, specifying the device in a lazy way will result in 
```
print(device.index)
>>> None
```

and then this line (L223)
```
device.type == "cuda" and device.index >= torch.cuda.device_count():
```
will fail. With this fix, it will no longer fail.

### DIFF
--- a/neuralpredictors/training.py
+++ b/neuralpredictors/training.py
@@ -218,9 +218,8 @@ def device_state(model, device):
     """
     # infer the original device based on the device the first parameter is placed on
     original_device = next(model.parameters()).device
-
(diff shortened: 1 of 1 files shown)


ArneNx:
Could you elaborate on what problem this fixes? I.e. why is this necessary?

cblessing24 (approved):

PR #125, opened 2021-04-13
Adds 'multiplex' argument to the FullGaussian2d readout.

KonstantinWilleke:
This new arguments multiplexes the neuronal activity, such that each neuron is copied to all locations, and the output at each individual location is returned.
Thus, the returned output matrix is no longer the shape of (Batch, Neurons), but (Batch, Neurons * Locations). This argument is useful when using the reconstruction or MEI pipeline.

### DIFF
--- a/neuralpredictors/layers/readouts.py
+++ b/neuralpredictors/layers/readouts.py
@@ -944,7 +944,7 @@ class FullGaussian2d(nn.Module):
         self.register_buffer("grid_sharing_index", torch.from_numpy(sharing_idx))
         self._shared_grid = True
 
-    def forward(self, x, sample=None, shift=None, out_idx=None, multiplex=False, crop_edge_px=None, collapse=True):
+    def forward(self, x, sample=None, shift=None, out_idx=None, multiplex=False, crop_edge_px=None, collapse=True, **kwargs):
         """
         Propagates the input forwards through the readout
         Args:
@@ -999,7 +999,7 @@ class FullGaussian2d(nn.Module):
             y = F.grid_sample(x, grid, align_corners=self.align_corners)
             y = (y.squeeze(-1) * feat).sum(1).view(N, outdims)
         else:
-            y = torch.einsum("ncwh,nco->nwho", x, feat)
+            y = torch.einsum("ncwh,uco->nwho", x, feat)
 
         if self.bias is not None:
             y = y + bias
@@ -1009,6 +1009,9 @@ class FullGaussian2d(nn.Module):
                 y = y[:, crop_edge_px:-crop_edge_px, crop_edge_px:-crop_edge_px, :]
             if collapse:
                 y = y.reshape(N, -1)
+            if not collapse:
+                # reshape responses into (N, Outdims, h, w)
+                y = y.permute(0, 3, 1, 2)
 
         return y
 
--- a/neuralpredictors/layers/readouts.py
+++ b/neuralpredictors/layers/readouts.py
@@ -944,7 +944,7 @@ class FullGaussian2d(nn.Module):
         self.register_buffer("grid_sharing_index", torch.from_numpy(sharing_idx))
         self._shared_grid = True
 
-    def forward(self, x, sample=None, shift=None, out_idx=None, multiplex=None, crop_edge_px=None):
+    def forward(self, x, sample=None, shift=None, out_idx=None, multiplex=False, crop_edge_px=None, collapse=True):
         """
         Propagates the input forwards through the readout
         Args:
@@ -995,7 +995,7 @@ class FullGaussian2d(nn.Module):
         if shift is not None:
             grid = grid + shift[:, None, None, :]
 
-        if multiplex is not True:
+        if not multiplex:
             y = F.grid_sample(x, grid, align_corners=self.align_corners)
             y = (y.squeeze(-1) * feat).sum(1).view(N, outdims)
         else:
@@ -1004,12 +1004,12 @@ class FullGaussian2d(nn.Module):
         if self.bias is not None:
             y = y + bias
 
-        if multiplex is True:
-            y = (
-                y.reshape(N, -1)
-                if crop_edge_px is None
-                else y[:, crop_edge_px:-crop_edge_px, crop_edge_px:-crop_edge_px, :].reshape(N, -1)
-            )
+        if multiplex:
+            if crop_edge_px:
+                y = y[:, crop_edge_px:-crop_edge_px, crop_edge_px:-crop_edge_px, :]
+            if collapse:
+                y = y.reshape(N, -1)
+
         return y
 
     def __repr__(self):
--- a/neuralpredictors/layers/readouts.py
+++ b/neuralpredictors/layers/readouts.py
@@ -1019,9 +1019,10 @@ class FullGaussian2d(nn.Module):
         if multiplex:
             if crop_edge_px:
                 y = y[:, crop_edge_px:-crop_edge_px, crop_edge_px:-crop_edge_px, :]
+
             if collapse:
                 y = y.reshape(N, -1)
-            if not collapse:
+            else:
                 # reshape responses into (N, Outdims, h, w)
                 y = y.permute(0, 3, 1, 2)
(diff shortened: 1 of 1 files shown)


neuralpredictors/layers/readouts.py
```diff
@@ -944,7 +944,7 @@ def initialize_shared_grid(self, match_ids=None, shared_grid=None):
         self.register_buffer("grid_sharing_index", torch.from_numpy(sharing_idx))
         self._shared_grid = True
 
-    def forward(self, x, sample=None, shift=None, out_idx=None):
+    def forward(self, x, sample=None, shift=None, out_idx=None, multiplex=None, crop_edge_px=None):
```
eywalker, line 947: Let `multiplex` default to `False`, and let's also add a new argument to control whether the output gets collapsed (e.g. `collapse=True`)

neuralpredictors/layers/readouts.py
```diff
@@ -988,11 +995,21 @@ def forward(self, x, sample=None, shift=None, out_idx=None):
         if shift is not None:
             grid = grid + shift[:, None, None, :]
 
-        y = F.grid_sample(x, grid, align_corners=self.align_corners)
-        y = (y.squeeze(-1) * feat).sum(1).view(N, outdims)
+        if multiplex is not True:
```
eywalker, line 998:
```python
if not multiplex:
```

neuralpredictors/layers/readouts.py
```diff
@@ -988,11 +995,21 @@ def forward(self, x, sample=None, shift=None, out_idx=None):
...
+            y = F.grid_sample(x, grid, align_corners=self.align_corners)
+            y = (y.squeeze(-1) * feat).sum(1).view(N, outdims)
+        else:
+            y = torch.einsum("ncwh,nco->nwho", x, feat)
 
         if self.bias is not None:
             y = y + bias
+
+        if multiplex is True:
```
eywalker, line 1007:
```python
if multiplex:
```

neuralpredictors/layers/readouts.py
```diff
@@ -988,11 +995,21 @@ def forward(self, x, sample=None, shift=None, out_idx=None):
...
+            y = (
+                y.reshape(N, -1)
```
eywalker, line 1009: in light of the new `collapse` argument, I would write it like:

```python
if crop_edge_px is not None:
    y = y[:, crop_edge_px:-crop_edge_px, crop_edge_px:-crop_edge_px, :]

if collapse:
    y = y.reshape(N, -1)
```

eywalker (changes requested):

eywalker (approved):

neuralpredictors/layers/readouts.py
```diff
@@ -956,6 +958,13 @@ def forward(self, x, sample=None, shift=None, out_idx=None):
                            if sample is True/False, overrides the model_state (i.e training or eval) and does as instructed
             shift (bool): shifts the location of the grid (from eye-tracking data)
             out_idx (bool): index of neurons to be predicted
+            multiplex (bool): if True, the neurons do not readout from their grid position,
+                            but from every position, i.e. every pixel, instead. Thus, each neuron is effectively copied
+                            to all positions in the image. The neuronal activity output matrix is then no longer
+                            (Batch, Neurons), but (Batch, Neurons x locations)
+            crop_edge_px (int): When multiplex is True, the response to all pixels that are cropped. Cropping
```
eywalker, line 965: could you include description about `collapse` and also update `multiplex` argument accordingly (about the shape of the output)

eywalker (changes requested):

neuralpredictors/layers/readouts.py
KonstantinWilleke: Thanks, I've addressed it!

neuralpredictors/layers/readouts.py
```diff
@@ -956,6 +958,19 @@ def forward(self, x, sample=None, shift=None, out_idx=None):
...
+                            will be symmetric by n pixels as specified by this argument.
+            collapse (bool): This argument can be used to modify the output shape of the readout when multiplex is
+                            set to True.
+                            By default (collapse=True), the readout returns the output in the shape of
+                            (batch, neurons x locations). When collapse=False, the returned shape will be:
+                            (batch, neurons, height x width)
+
 
         Returns:
             y: neuronal activity
```
PPierzc, line 976: Since multiplex and collapse modify what is returned it would be good to also have that mentioned in the return docstring.

Maybe something like this:
```suggestion
            y: neuronal activity. shape: default (batch, neurons), if multiplex=True and collapse=False (batch, neurons x locations), if multiplex=True and collapse=True (batch, neurons, height x width)
```

neuralpredictors/layers/readouts.py
```diff
@@ -988,11 +1003,24 @@ def forward(self, x, sample=None, shift=None, out_idx=None):
...
 
         if self.bias is not None:
             y = y + bias
+
+        if multiplex:
+            if crop_edge_px:
+                y = y[:, crop_edge_px:-crop_edge_px, crop_edge_px:-crop_edge_px, :]
+            if collapse:
+                y = y.reshape(N, -1)
+            if not collapse:
```
PPierzc, line 1020:
```suggestion
           else:
```
Is there a reason for splitting it into two checks?

neuralpredictors/layers/readouts.py
```diff
@@ -988,11 +1003,24 @@ def forward(self, x, sample=None, shift=None, out_idx=None):
         if shift is not None:
             grid = grid + shift[:, None, None, :]
 
-        y = F.grid_sample(x, grid, align_corners=self.align_corners)
-        y = (y.squeeze(-1) * feat).sum(1).view(N, outdims)
+        if not multiplex:
```
PPierzc, line 1006:
```suggestion
       if multiplex:
            y = torch.einsum("ncwh,uco->nwho", x, feat)
        else:
            y = F.grid_sample(x, grid, align_corners=self.align_corners)
            y = (y.squeeze(-1) * feat).sum(1).view(N, outdims)
```

I think checking negatives and having the else clause makes it a bit more difficult to follow.

neuralpredictors/layers/readouts.py
```diff
@@ -988,11 +1003,24 @@ def forward(self, x, sample=None, shift=None, out_idx=None):
...
+            y = (y.squeeze(-1) * feat).sum(1).view(N, outdims)
+        else:
+            y = torch.einsum("ncwh,uco->nwho", x, feat)
 
         if self.bias is not None:
             y = y + bias
+
+        if multiplex:
+            if crop_edge_px:
+                y = y[:, crop_edge_px:-crop_edge_px, crop_edge_px:-crop_edge_px, :]
```
PPierzc, line 1017:
```suggestion
                y = y[:, crop_edge_px:-crop_edge_px, crop_edge_px:-crop_edge_px, :]
                
```
I prefer to separate ideas/thoughts visually by adding a blank line between them.

neuralpredictors/layers/readouts.py
```diff
@@ -988,11 +1003,24 @@ def forward(self, x, sample=None, shift=None, out_idx=None):
         if shift is not None:
             grid = grid + shift[:, None, None, :]
 
-        y = F.grid_sample(x, grid, align_corners=self.align_corners)
-        y = (y.squeeze(-1) * feat).sum(1).view(N, outdims)
+        if not multiplex:
+            y = F.grid_sample(x, grid, align_corners=self.align_corners)
+            y = (y.squeeze(-1) * feat).sum(1).view(N, outdims)
+        else:
+            y = torch.einsum("ncwh,uco->nwho", x, feat)
```
PPierzc, line 1010: Maybe it is just me, but I can never immediately say what the einsum exactly is supposed to do without extensively thinking about it. You could add a small comment explaining what the einsum is supposed to do.

PPierzc (changes requested):

PPierzc (approved):

eywalker (approved):

PR #126, opened 2021-04-15
bugfix for device_state when device is "cuda"

KonstantinWilleke:


neuralpredictors/training/context_managers.py
```diff
@@ -46,7 +46,8 @@ def device_state(model, device):
     original_device = next(model.parameters()).device
 
     # create device spec
-    device = torch.device(device)
+    # if device is simply "cuda", then device.index will evaluate to one, and the if statement will error out.
+    device = torch.device("cuda:0") if device == "cuda" else torch.device(device)
     if device.type == "cuda" and device.index >= torch.cuda.device_count():
```
cblessing24, line 51: You could also do `getattr(device, "index", 0)` here instead of changing how `device` is created above. Not sure if that would be preferable.

cblessing24 (approved):

PR #135, opened 2021-06-23
Encoder

kklurz:
Create a standard encoder class that works with single-parameter distributions (i.e. Poisson). As a result:
- tiny changes were introduced to the MLP shifter class to keep the encoder forward function definition generic, and
- **[IMPORTANT]** the trainer is now expected to unpack the Datapoint named tuple returned by the dataloader into `input` as a positional argument and other tensors (e.g. `targets`, `behavior`, etc.) as optional arguments for the forward function of the encoder.


neuralpredictors/layers/encoders/poisson.py
```diff
@@ -0,0 +1,34 @@
+from torch import nn
+
+
+class PoissonEncoder(nn.Module):
```
KonstantinWilleke, line 4: Now that I think of it - I think it could be nice to have a more abstract base class which is just called Encoder, without the fancy stuff. So just core/readout/elu_offset. Maybe sitting in base.py. Then it's a consistent structure.
The PoissonEncoder can then inherit from it.

neuralpredictors/layers/encoders/poisson.py
```diff
@@ -0,0 +1,34 @@
...
+        self.offset = elu_offset
+
+    def forward(
+        self,
+        inputs,
+        targets=None,
+        data_key=None,
+        behavior=None,
+        pupil_center=None,
+        trial_idx=None,
```
KonstantinWilleke, line 19: The trial_idx is a special variant, which is not really needed in most cases. I think it's best omit it here, because in the default shifter, that argument isn't needed and will throw an error in the default shifers.

neuralpredictors/layers/encoders/poisson.py
```diff
@@ -0,0 +1,34 @@
+from torch import nn
```
KonstantinWilleke, line 1: Could you add a small text for the documentation in the __init__ as well?

KonstantinWilleke (changes requested):

neuralpredictors/layers/encoders/poisson.py
mohammadbashiri: Which shifter is the default shifter? We tried to change the shifters such that they all accept it, but don't have to use it.

neuralpredictors/layers/encoders/poisson.py
kklurz: Mh, but what would we outsource to that base class? Everything that the PoissonEncoder has, is already very specific

neuralpredictors/layers/encoders/poisson.py
kklurz: Yes, Mohammad and I thought that it would be better to let the encoder always pass everything (i.e. pupil_center and trial_idx) and then the shifter itself needs to deal with it. That's why we changed all shifters to accept trial_idx. Do you prefer the encoder dealing with it @KonstantinWilleke?

neuralpredictors/layers/encoders/poisson.py
KonstantinWilleke: At least the init can be inherited. So that the base class takes care of initializing core/readout/offset. And with the `super().__init__`, the base class always has that job. I think the overall advantage is that it is very easy to understand what the base class is doing, if the forward is also so simple.
Something like:
```

x = self.core(inputs)
x = self.readout(x, data_key=data_key, shift=shift, **kwargs)
return nn.functional.elu(x + self.offset) + 1
```

KonstantinWilleke: then the poisson class would overwrite the forward, but it's clean and understandable at first glance.

neuralpredictors/layers/encoders/poisson.py
KonstantinWilleke: The default shifter is the `MLP`.  It takes one input arg. And the trial idx is used in my shifter like this: I stack the pupil center and trial index, and then pass it as one argument. But that's not the "standard" case of dealing with the shifter. The normal case would be just to pass the eye position to the shifter. So I think the trial_idx is not general enough to be part of neuralpredictors.

mohammadbashiri: If we remove the `trial_index` from the encoder, then you will not be able to use the encoder here out of the box and would need to inherit and modify forward function, right? So far we only have two shifter classes, so I think it would not hurt to pass the trial_index as an optional argument and potentially use them in the shifter. You are using it because of the problem we had with the camera movement, right? and that might happen again in the future.

KonstantinWilleke: Yes true, I'd need to modify the forward function then. But the shifter itself can not deal, at least not at the moment, with more than 1 input. So my encoder does the stacking of pupil_center and trial_idx. And adding this stacking as the default for the poisson encoder might be a bit much, but it possibly also wouldn't hurt. 
The only downside I could imagine is this: The trial_idx might be used by some other mechanic as well. Maybe for behavior, or some other hypothetical case where something changes across the experiment. So if a trial_idx is present, stacking it by default with the eye data might lead to problems.

KonstantinWilleke: Ahhh now I see, sorry, you have modified the shifters!

mohammadbashiri: The stacking is a shifter-specific thing, and that is why it should be dealt with in the shifter itself.
Yes exactly, we also changed the shifters. So the encoder and shifter in this PR should work together out of the box, theoretically.

KonstantinWilleke: thanks, yes I like it!

neuralpredictors/layers/shifters/static_affine.py
```diff
@@ -21,6 +21,8 @@ def __init__(self, input_channels=2, output_channels=2, bias=True):
         super().__init__(input_channels, output_channels, bias=bias)
 
     def forward(self, x, trial_idx=None):
+        if trial_idx is not None:
+            raise ValueError("This shifter network does not support trial indexing. Set trial_idx to None")
```
mohammadbashiri, line 25: I am not so sure about the error here. I think it's unnecessary since the user would probably choose a specific shifter depending on whether they want to make use of `trial_index` or not. Having `trial_index` as input argument for all shifters is just a convenience and a convention (just like the `shift` argument for the readouts). Also, the error has potential to become a hassle later on. Here is an example:
Imagine we have two models, where each one of them use a different shifter (one that uses `trial_index`, one doesn't). With this error, we would need two dataloaders (one that returns `trial_idx` and one that does not). Without it, we can just share one dataloader between the two models and have them both work smoothly (at least as far as the shifter is concerned)

mohammadbashiri (changes requested):

neuralpredictors/layers/encoders/poisson.py
kklurz: ok, so I resolve this?

neuralpredictors/layers/encoders/poisson.py
kklurz: But that code you sent there is also specific to the PoissonEncoder, it has the `elu + 1` which is designed to represent a firing rate, the parameter of the poisson distribution. So the `forward` can not be outsourced, it always depends on the distribution. 
And the `__init__` would not gain anything from outsourcing to a base class codewise because you would need to split up the arguments and pass some of them to `super().__init__()` and others to init itself... That's just as much code... I guess one culd make an argument that it is slightly easier to understand. But then again it's really not much and one needs to look up a completely new file just to see 2 lines of code where the core and readout arguments are turned into a class attribute... 
Not sure, any opinions?

KonstantinWilleke: Yea you are right, the `+ 1` is a poisson thing. I still like the base class. I'm having the same `__init__` issue with the cores. With one base class having the init, and the children have very similar `__inits__`s.

I'm solving it by doing:
in the base class:
`def __init__(arg1, arg2, arg3=1):`...

and then in the children:
```
def __init__(*args, arg4=100, arg5=200, **kwargs):
   super().__init__(*args, **kwargs)
   self.arg4=arg4
   self.arg5=arg5
   ```

Comment (KonstantinWilleke) [reply neuralpredictors/layers/encoders/poisson.py:4]: Then the base class still has its role in the init, and the children will only initialize what they add on top.

Comment (kklurz) [reply neuralpredictors/layers/encoders/poisson.py]: So in the meeting we agreed on not having a base class for the moment. I'll close this conversation for now

File: neuralpredictors/layers/shifters/static_affine.py (hunk 4 above)
Comment (KonstantinWilleke) [reply neuralpredictors/layers/shifters/static_affine.py:25]: To summarize this issue as we talked about it in the code-review session: I agree that an error or warning is over the top, and I'd suggest to use **kwargs in the shifter instead.

File: neuralpredictors/layers/shifters/mlp.py [hunk 5]
```diff
@@ -37,11 +38,14 @@ def initialize(self):
         for linear_layer in [p for p in self.parameters() if isinstance(p, nn.Linear)]:
             xavier_normal(linear_layer.weight)
 
-    def forward(self, input):
-        return self.mlp(input)
+    def forward(self, pupil_center, trial_idx=None):
+        if trial_idx is not None:
+            if self.mlp[0].in_features == 3:
```
cblessing24, line 43: Please comment on this line.

neuralpredictors/layers/shifters/static_affine.py
cblessing24: In the code review meeting today we decided to raise a warning here if `trial_idx is not None`.

cblessing24 (changes requested):

cblessing24 (approved):
Looks dandy. :+1:

mohammadbashiri (approved):

KonstantinWilleke (approved):

PR #138, opened 2021-11-17
Last fixes before publication

kklurz:
Fixes include:

- numpy.typing (used in np_functions.py) is only available from numpy>=1.20
- imports in __init__.py
- The naming of conv layers in the core was not backwards compatible. Specifically, the depth-separable layer used to be called "ds_conv" and is called "conv" at the moment. The PR changes that. Additionally, the PR also changes the naming of the attention conv and hermite conv. I am not sure, if the new naming there is backwards compatible. PLEASE CHECK


KonstantinWilleke:
Thanks a lot for catching these errors!

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -146,10 +146,14 @@ def __init__(
         self.linear = linear
 
         if depth_separable:
+            self.conv_layer_name = "ds_conv"
             self.ConvLayer = DepthSeparableConv2d
         elif attention_conv:
+            # TODO: check if name attention_conv is backwards compatible
+            self.conv_layer_name = "attention_conv"
```
KonstantinWilleke, line 153: So far, the attention conv was only used by me in the SECore2d as far as I know. There, I didn't give the layer a specific name, but just `conv`. 
So to make it backwards compatible, I think it would be best to also just name it `conv`.

https://github.com/sinzlab/nnvision/blob/c9e69c308248f50c4362254eae68f70414570076/nnvision/models/cores.py#L200

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -332,7 +336,7 @@ def set_batchnorm_type(self):
 
     def add_first_layer(self):
         layer = OrderedDict()
-        layer["conv"] = HermiteConv2D(
+        layer["hermite_conv"] = HermiteConv2D(
```
KonstantinWilleke, line 339: same with the HermiteConv. Previously it was also just named `conv`.

https://github.com/sinzlab/neuralpredictors/blob/862198b1790c3a1779a87930dbbac8ef05370e0c/neuralpredictors/layers/cores.py#L360

KonstantinWilleke (changes requested):

KonstantinWilleke:
Looking back, it was probably a mistake of mine to only rename the DepthSeparableConv into `ds_conv` while all others are just conv. In order to make it truly backwards compatible though, it might be best to keep it.

kklurz:
OK, so we have to make a compromise between being backwards compatible and making our code logical and thorough. You said that you are the only one who ever used the attention conv? Were you also the only one using the Hermite conv? If I remember correctly, they are both quite new cores. How easy/hard would it be for you if we decided to do the naming properly? Would you lose a lot of your previous work?

kklurz:
There is another problem we need to address:

The new FiringRateEncoder takes these arguments: `core, readout, shifter=None, modulator=None, elu_offset=0.0`, where shifter and modulator used to not be there. If people used to instantiate the Encoder class with positional arguments, i.e. `model = Encoder(core, readout, elu_offset)`, it will now assume that the value for `elu_offset` is a shifter which will **not** throw and error but cause problems further down the line of course.

We can either change the order of arguments to `core, readout, elu_offset=0.0, shifter=None, modulator=None` or we can include some checks in the code

neuralpredictors/layers/cores/conv2d.py
kklurz: We agreed on keeping the more descriptive names because the module is relatively new and not much code would break because of the missing backwards compatibility

KonstantinWilleke (approved):

PR #140, opened 2021-11-18
Fix bug in MultiReadoutSharedParametersBase

KonstantinWilleke:
The MultiReadoutSharedParametersBase is used per default when initializing the MultipleGaussian2d.
It assumes that the kwargs of the shared parameters, such as `share_transform` are passed to the MultiReadout. But this is not necessarily the case (for example, when a normal MultipleGaussian is initialized).
I therefore changed the `del` functions to `pop` methods.

If we make MultiReadoutSharedParametersBase to only work with readout that should have parameters that are shared, we could do something like this in the init instead, and get around the problem (and introduce new problems instead):

```
class MultipleFullGaussian2d(MultiReadoutBase):
    _base_readout = FullGaussian2d
   
class MultipleFullGaussian2dShared(MultiReadoutSharedParametersBase):
    _base_readout = FullGaussian2d
```

### DIFF
--- a/neuralpredictors/layers/readouts/multi_readout.py
+++ b/neuralpredictors/layers/readouts/multi_readout.py
@@ -105,7 +105,10 @@ class MultiReadoutSharedParametersBase(MultiReadoutBase):
                     "match_ids": readout_kwargs["shared_match_ids"][data_key],
                     "shared_grid": None if i == 0 else self[first_data_key].shared_grid,
                 }
-            _ = [readout_kwargs.pop(kwarg, None) for kwarg in ["share_transform", "share_grid", "grid_mean_predictor_type"]]
+            _ = [
+                readout_kwargs.pop(kwarg, None)
+                for kwarg in ["share_transform", "share_grid", "grid_mean_predictor_type"]
+            ]
 
         if "share_features" in readout_kwargs:
             if readout_kwargs["share_features"]:
--- a/neuralpredictors/layers/readouts/multi_readout.py
+++ b/neuralpredictors/layers/readouts/multi_readout.py
@@ -91,22 +91,22 @@ class MultiReadoutSharedParametersBase(MultiReadoutBase):
         i,
         data_key,
         first_data_key,
+        grid_mean_predictor=None,
+        grid_mean_predictor_type=None,
         share_transform=None,
         share_grid=None,
-        grid_mean_predictor=None,
         share_features=None,
         **kwargs
     ):
         readout_kwargs = kwargs.copy()
 
         if grid_mean_predictor:
-            if readout_kwargs["grid_mean_predictor_type"] == "cortex":
+            if grid_mean_predictor_type == "cortex":
                 readout_kwargs["source_grid"] = readout_kwargs["source_grids"][data_key]
+                readout_kwargs["grid_mean_predictor"] = grid_mean_predictor
             else:
-                raise KeyError(
-                    "grid mean predictor {} does not exist".format(readout_kwargs["grid_mean_predictor_type"])
-                )
-            if share_transform is not None:
+                raise KeyError("grid mean predictor {} does not exist".format(grid_mean_predictor_type))
+            if share_transform:
                 readout_kwargs["shared_transform"] = None if i == 0 else self[first_data_key].mu_transform
 
         elif share_grid:
--- a/neuralpredictors/layers/readouts/multi_readout.py
+++ b/neuralpredictors/layers/readouts/multi_readout.py
@@ -86,37 +86,40 @@ class MultiReadoutSharedParametersBase(MultiReadoutBase):
     For more information on which parameters can be shared, refer for example to the FullGaussian2d readout
     """
 
-    def prepare_readout_kwargs(self, i, data_key, first_data_key, **kwargs):
+    def prepare_readout_kwargs(
+        self,
+        i,
+        data_key,
+        first_data_key,
+        share_transform=None,
+        share_grid=None,
+        grid_mean_predictor=None,
+        share_features=None,
+        **kwargs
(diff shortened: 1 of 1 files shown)


kklurz:
Good catch! I'm not sure though if your suggested changes would work. Did you test this? 

If I understand it right, a line like [this](https://github.com/sinzlab/neuralpredictors/blob/05636486a39175af52d86b00f01c9d1de40213b6/neuralpredictors/layers/readouts/multi_readout.py#L100) would still break, if the key is not in the dictionary `readout_kwargs`. 

I can think of 2 solutions:

Nested if statements: 

```
if key in readout_kwargs:
    if readout_kwargs[key]:
        ....
```

Or we add the keys in the beginning:

```
readout_kwargs[key] = False if key not in readout_kwargs
```

KonstantinWilleke:
I chose to do the if statements like
`if grid_mean_predictor:`... So that both `None` and `False` would skip these steps.
I've tested it with my normal gaussian model and it works, but I havent tested it with a shared readout (because I've never used it so far).

kklurz:
I don't think this would work... 

In the original version, the keys that are deleted are:
```
del readout_kwargs["share_transform"]
del readout_kwargs["share_grid"]
del readout_kwargs["grid_mean_predictor_type"]
...
del readout_kwargs["share_features"]
```

In your PR, the key `grid_mean_predictor` would be "deleted" instead of `grid_mean_predictor_type`.

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -86,37 +86,40 @@ class MultiReadoutSharedParametersBase(MultiReadoutBase):
...
+        if grid_mean_predictor:
+            if readout_kwargs["grid_mean_predictor_type"] == "cortex":
+                readout_kwargs["source_grid"] = readout_kwargs["source_grids"][data_key]
             else:
-                readout_kwargs["shared_features"] = None
-            del readout_kwargs["share_features"]
+                raise KeyError(
+                    "grid mean predictor {} does not exist".format(readout_kwargs["grid_mean_predictor_type"])
+                )
+            if share_transform is not None:
```
kklurz, line 109: Why do you compare to `None` and not do `if share_transform` like with the other keys?

neuralpredictors/layers/readouts/multi_readout.py
```diff
@@ -86,37 +86,40 @@ class MultiReadoutSharedParametersBase(MultiReadoutBase):
     For more information on which parameters can be shared, refer for example to the FullGaussian2d readout
     """
 
-    def prepare_readout_kwargs(self, i, data_key, first_data_key, **kwargs):
+    def prepare_readout_kwargs(
+        self,
+        i,
+        data_key,
+        first_data_key,
+        share_transform=None,
```
kklurz, line 94: I think logic-wise, it only makes sense to let `share_transform`, `share_grid` and `share_features` default to boolean `False`, while `grid_mean_predictor` should default to `None`. But as explained in my main comment, `grid_mean_predictor` needs to be part of the `readout_kwargs` that ultimately get returned by the function

kklurz (changes requested):

neuralpredictors/layers/readouts/multi_readout.py
KonstantinWilleke: yes good point, I corrected it!

KonstantinWilleke:
Yes I see all of your points from the main comment. I've addressem them now by explicitly adding both grid_mean_predictor arguments. And then adding `grid_mean_predictor` to the `readout_kwargs` if it is not None.

neuralpredictors/layers/readouts/multi_readout.py
KonstantinWilleke: I think it is cleaner to have them default to None, when they are not expected to be passed in to the function.

kklurz: Mh, I disagree... Their only purpose is this one function and in there, they have a clear meaning and task, which is boolean. What does it mean if someone passes in a list for `share_transform` and the list has no meaning but it is treated as a `True`? Whenever people look up what arguments they could give to their MultiReadout with shared parameters, this is the place where they will look and they should see what type these arguments are. This being said, we should also have a docstring there.

PPierzc: I agree with @kklurz on this. `share_transform` etc. are bools and the default mode is that you do not share. If I would see `None` as a default that would suggest that 1) the default value is mutable, 2) I have to always specify whether I want to share or not and think about it, which seems like it's not the point that we are trying to make with this function.

mohammadbashiri: I think it depends on what kinds of values they are expected to have (i.e. what type they are expected to be). As far as I understand `share_transform`, `share_grid`, and `share_features` are expected to be boolean, and `grid_mean_predictor` can be a string. So my suggestion would be defaulting the booleans to some default boolean and the string to None - similar to what @kklurz mentioned.

KonstantinWilleke: Ok thanks! I'm convinced here. I'll turn all but `grid_mean_predictor` into bools!

cblessing24: I agree that boolean parameters should be `False` by default and the string one should be `None`. Also right now the `if` statement on line 109 is `True` if `share_transform` is `False`. Is that intended?

cblessing24: Nevermind. I was looking at an older version...

mohammadbashiri (approved):

kklurz (approved):

PR #145, opened 2021-12-13
Annotate Base Readout

cblessing24:
- Annotated base readout module
- Remove unnecessary mypy ignores from attention readout module

### DIFF
--- a/mypy_files.txt
+++ b/mypy_files.txt
@@ -1,1 +1,2 @@
 neuralpredictors/layers/readouts/attention.py
+neuralpredictors/layers/readouts/base.py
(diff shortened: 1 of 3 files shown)


MaxFBurg (approved):
Looks good to me. I guess the `# type: ignore`s might be resolved later?

MaxFBurg (approved):

cblessing24:
> Looks good to me. I guess the `# type: ignore`s might be resolved later?

Yes, these are necessary when dealing with types from other modules that have not been annotated yet. Once that has happened, the ignores can be removed.

PR #150, opened 2021-12-16
merge bug fixes of main into dev branch

KonstantinWilleke:


neuralpredictors/layers/cores/conv2d.py
```diff
@@ -249,6 +249,10 @@ def group_sparsity(self):
         Sparsity regularization on the filters of all the conv2d layers except the first one.
         """
         ret = 0
+
+        # return zero for depth separable convolutions
+        if not hasattr(self.features[1], "conv"):
+            return ret
```
mohammadbashiri, line 255: Could you please add a warning as well to indicate that the sparsity regularization is not applied? At the moment it is being silently ignored, which may be misleading when training models. I am not sure if this is the best place to put the warning tho. Maybe best is at initialization: check whether conv type is not `conv` and if `hidden_gamma` is not 0, then we should have a warning.

mohammadbashiri (changes requested):

mohammadbashiri:
Also black is failing.

kklurz (commented):
I don't have anything to add except the two issues Mohammad already raised

neuralpredictors/layers/cores/conv2d.py
KonstantinWilleke: Yes, good points! I added it to the init and printed a warning

mohammadbashiri (approved):

neuralpredictors/layers/cores/conv2d.py
```diff
@@ -148,14 +148,21 @@ def __init__(
...
+            self.ignore_group_sparsity = True
         else:
             self.conv_layer_name = "conv"
             self.ConvLayer = nn.Conv2d
+            self.ignore_group_sparsity = False
 
+        if self.ignore_group_sparsity:
+            warnings.warn(
+                "group sparsity can not be calculated for the requested conv type. Hidden channels will not be regularized and gamma_hidden is ignored."
+            )
```
kklurz, line 165: I think this warning should only be thrown if a regularization of the hidden channels was actually intended. Otherwise this will be quite annoying and it will also make the user wonder about things they don't need to wonder about. 
I suggest `if self.ignore_group_sparsity and gamma_hidden > 0`

kklurz (changes requested):

neuralpredictors/layers/cores/conv2d.py
KonstantinWilleke: Yes, good point. modifying now!

kklurz (approved):

PR #152, opened 2022-01-07
Make encoder forward more forgiving

KonstantinWilleke:


neuralpredictors/layers/encoders/firing_rate.py
```diff
@@ -22,7 +22,6 @@ def __init__(self, core, readout, *, shifter=None, modulator=None, elu_offset=0.
 
     def forward(
         self,
-        inputs,
```
mohammadbashiri, line 25: Couldn't this be taken care of on the challenge repo side? Because this change will be going back to what we had before.

mohammadbashiri: My suggestion would be to leave neuralpredictors untouched, and create a modified Encoder in the challenge repo (i.e. inheriting from the [FiringRateEncoder](https://github.com/sinzlab/neuralpredictors/blob/2279feb134a7a12a58d2f20c670febbd94e127bd/neuralpredictors/layers/encoders/firing_rate.py#L4) with [modified `forward` function](https://github.com/mohammadbashiri/neural-prediction-challenge/blob/2f99cd7c815478484239dde4a400c0ac37a2f623/cascade/models/models.py#L19)). I already implemented this suggestion here and tested the pipeline -> https://github.com/sinzlab/neural-prediction-challenge/pull/29.

KonstantinWilleke: Ok I agree here to leave neuralpredictors untouched for now. But I still think that ideally the default encoder should be applicable to most use-cases. But we can discuss this at a later point in a code review!
Will make an issue and close

PR #163, opened 2022-02-04
Fix final non-linearity behavior in Stacked2dCore

cblessing24:
- Test non-linearity behavior in Stacked2dCore
- Fix failing non-linearity tests from 10a4534
- Add test_conv2d.py to mypy checked files

### DIFF
--- a/neuralpredictors/layers/cores/conv2d.py
+++ b/neuralpredictors/layers/cores/conv2d.py
@@ -182,8 +182,11 @@ class Stacked2dCore(Core, nn.Module):
                     layer["scale"] = self.scale_layer_cls(self.hidden_channels)
 
     def add_activation(self, layer):
-        if (self.num_layers > 1 or self.final_nonlinearity) and not self.linear:
-            layer["nonlin"] = AdaptiveELU(self.elu_xshift, self.elu_yshift)
+        if self.linear:
+            return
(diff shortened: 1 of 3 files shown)


neuralpredictors/layers/cores/conv2d.py
```diff
@@ -182,8 +182,11 @@ def add_bn_layer(self, layer):
                     layer["scale"] = self.scale_layer_cls(self.hidden_channels)
 
     def add_activation(self, layer):
-        if (self.num_layers > 1 or self.final_nonlinearity) and not self.linear:
-            layer["nonlin"] = AdaptiveELU(self.elu_xshift, self.elu_yshift)
+        if self.linear:
+            return
+        if len(self.features) == self.num_layers - 1 and not self.final_nonlinearity:
+            return
+        layer["nonlin"] = AdaptiveELU(self.elu_xshift, self.elu_yshift)
```
ArneNx, line 189: I would personally prefer to make the condition more explicit about when a nonlinearity will be added:
```python
if self.linear:
    return
if len(self.features) < self.num_layers - 1 or self.final_nonlinearity:
    layer["nonlin"] = AdaptiveELU(self.elu_xshift, self.elu_yshift)
```

cblessing24: I changed that.

ArneNx (approved):

PR #164, opened 2022-02-10
Tracker update

ArneNx:
Changes tracker to allow log reset after each epoch


ArneNx:
I added this option to reset the tracker, instead of appending a new epoch, because it is sometimes unnecessary to save the entire training process. 
In cases that sometimes can have up to a million epochs, this change offers the option to reset the tracker for each new epoch instead of appending to the history of epochs.

KonstantinWilleke (approved):

PR #167, opened 2022-02-21
Add "How to contribute" section

cblessing24:
Add the "How to contribute" section which describes some preconditions
that pull requests need to fulfill.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,51 @@
 
 [Sinz Lab](https://sinzlab.org/) Neural System Identification Utilities for [PyTorch](https://pytorch.org/).
 
-## Running Tests
+## How to run the tests :test_tube:
 
 Clone this repository and run the following command from within the cloned repository to run all tests:
 
 ```bash
-docker-compose run pytest test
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,10 +8,51 @@
...
+
+### Tests
+
+Please make sure your changes pass the tests. Take a look at the [test running
+section](#how-to-run-the-tests-test_tube) for instructions on how to run them. Adding tests
+for new code is not mandatory but encouraged.
+
+### Code Style
+
+This project uses the [black](https://github.com/psf/black) code formatter. You
```
ArneNx, line 32: Specify the exact version here?

cblessing24: The version will be  the correct one if they run black using the compose service. I can add a reference to the `requirements-dev.txt.` file which contains the version? I'd like the versions to be specified in one central place.

ArneNx: Ah you're right. Running through docker-compose solves this.

ArneNx (approved):

Issue #176, opened 2022-09-12
Bug report - stride does not change for RotationEquivariant2dCore


pollytur:
I tried to change the stride for  RotationEquivariant2dCore, while it does not updates (checked with `model.stride`).
To reproduce: 
```
model = RotationEquivariant2dCore(..., stride=2)
print(model.stride) # should print 1
```

The issue I guess is that stride gets reinitialised by the default value during calling the` super.__init__` as stride is not passed there ([source code link](https://github.com/sinzlab/neuralpredictors/blob/main/neuralpredictors/layers/cores/conv2d.py#L378)).

The dirty fix (for someone who need it urgently) could be 

```
self.num_rotations = num_rotations
self.upsampling = upsampling
self.rot_eq_batch_norm = rot_eq_batch_norm
self.init_std = init_std
super().__init__(*args, **kwargs, input_regularizer=input_regularizer, stride=stride)
```

 or stride could also passes as a part of kwargs in further versions of the lib

mohammadbashiri:
Right. Any shared arg between `RotationEquivariant2dCore` and `Stacked2dCore` should be passed when calling `super().__init__()` and as far as I checked stride was the only one.

I think your suggestion is good. Alternatively, you could just remove it from the `RotationEquivariant2dCore` init args and also attributes - then it will be part of the `kwargs`. I would also suggest just calling the super in the beginning and then setting the other attributes. Would you mind creating a PR for this @pollytur?

pollytur:
Sure, I can make a PR for this with passing the argument then (like stated above if its fine)

As for calling the super in the beginning and then setting the other attributes - it will not work correctly without passing the stride as well as it initiates the network architecture with [add_subsequent_layers](https://github.com/sinzlab/neuralpredictors/blob/a761c1efa41c57eaea50075a6d7e4a56e74d1ca2/neuralpredictors/layers/cores/conv2d.py#L196), which already uses [stride](https://github.com/sinzlab/neuralpredictors/blob/a761c1efa41c57eaea50075a6d7e4a56e74d1ca2/neuralpredictors/layers/cores/conv2d.py#L263) and changing stride after `super.__init__()`  will not change the architecture

Issue #177, opened 2022-09-14
Question on strides


pollytur:
I am not sure if its a typo or it actually has some sense.

For `Stacked2dCore` in the function [add_first_layer](https://github.com/sinzlab/neuralpredictors/blob/main/neuralpredictors/layers/cores/conv2d.py#L236) the stride in not defined (the strike  value is not passed for the pytorch Conv2d layer), while it is passed for all other layers in [add_subsequent_layers](https://github.com/sinzlab/neuralpredictors/blob/main/neuralpredictors/layers/cores/conv2d.py#L249) and it is passed in [add_first_layer](https://github.com/sinzlab/neuralpredictors/blob/main/neuralpredictors/layers/cores/conv2d.py#L390) for `RotationEquivariant2dCore`. So, should the stride be passed for the layer0 as well?   (I assume its a typo)

Best, Polly

PR #178, opened 2022-09-15
[fix] equivariance stride now updates

pollytur:
#176

### DIFF
--- a/neuralpredictors/layers/cores/conv2d.py
+++ b/neuralpredictors/layers/cores/conv2d.py
@@ -371,11 +371,10 @@ class RotationEquivariant2dCore(Stacked2dCore, nn.Module):
             parameters and a BiasLayer will be added after the batch_norm layer.
         """
         self.num_rotations = num_rotations
-        self.stride = stride
         self.upsampling = upsampling
         self.rot_eq_batch_norm = rot_eq_batch_norm
         self.init_std = init_std
-        super().__init__(*args, **kwargs, input_regularizer=input_regularizer)
+        super().__init__(*args, **kwargs, input_regularizer=input_regularizer, stride=stride)
 
     def set_batchnorm_type(self):
         if not self.rot_eq_batch_norm:


cblessing24 (changes requested):
I think it would be better to pass all the arguments that `Stacked2dCore` needs via `*args` and `**kwargs` and don't even explicitly mention them in the signature of `__init__`. Then we could also simplify the docstring to say that these additional parameters will be passed to `Stacked2dCore`.

pollytur:
@cblessing24 I will add stride `*args` and `**kwargs` to but I do not think that we should also hide `input_regularizer` as in `Stacked2dCore` and `RotationEquivariant2dCore` there are different default values. 
Is it fine?

cblessing24:
Sounds good. Could you also update the docstring, i.e. remove all the arguments that are already described in the super class and add a line saying that `*args` and `**kwargs` will be passed to the super class please?

pollytur:
#177

pollytur:
yer, now all the arguments left are the ones specific for the `RotationEquivariant2dCore`

cblessing24:
Would you be willing to remove the duplication between the docstrings of the two classes?

cblessing24:
Could you format the docstring in the same way as `Stacked2dCore` please?

Something like:

cblessing24:
You are probably using a different version of black. The version used for the check is specified [here](https://github.com/sinzlab/neuralpredictors/blob/main/requirements-dev.txt). If you have docker-compose you can also format the code using the instructions from the [README](https://github.com/sinzlab/neuralpredictors#black).

mohammadbashiri:
Hey @pollytur. Thanks for the PR!

This change to the [`Stacked2dCore`](https://github.com/sinzlab/neuralpredictors/blob/1c64b9053fde65116f73e83c0ba99579de381398/neuralpredictors/layers/cores/conv2d.py#L242) is not backwards compatible. I'd suggest the following:

- in [`Stacked2dCore`](https://github.com/sinzlab/neuralpredictors/blob/1c64b9053fde65116f73e83c0ba99579de381398/neuralpredictors/layers/cores/conv2d.py#L28) add a new argument `input_stride=1` (default to 1)
- use `self.input_stride` to [initialize the first layer](https://github.com/sinzlab/neuralpredictors/blob/1c64b9053fde65116f73e83c0ba99579de381398/neuralpredictors/layers/cores/conv2d.py#L236-L245)
- when calling `super()` in the rotation-equivariant module just set both `stride` and `input_stride`

pollytur:
@mohammadbashiri could I ask why is it not backwards compatible? (the names of the arguments have not changed anywhere in the function names). If there are some old models which do not load now with the old config, this happens exactly because the stride was not updated, hence, the dimensions are different, which was the bug

mohammadbashiri:
> @mohammadbashiri could I ask why is it not backwards compatible? (the names of the arguments have not changed anywhere in the function names). If there are some old models which do not load now with the old config, this happens exactly because the stride was not updated, hence, the dimensions are different, which was the bug

What I am talking about is not the RotationEquivariantCore, but the Stacked2dCore module. It has a method to define the first layer. In the main branch at the moment this method [does not set the stride for the first layer](https://github.com/sinzlab/neuralpredictors/blob/1c64b9053fde65116f73e83c0ba99579de381398/neuralpredictors/layers/cores/conv2d.py#L236-L245) which means that the stride is set to 1 for the first layer and the `stride` argument was being used only for other layers. Your PR changes that [by setting the stride of the first layer using the `stride` arg](https://github.com/sinzlab/neuralpredictors/blob/1c64b9053fde65116f73e83c0ba99579de381398/neuralpredictors/layers/cores/conv2d.py#L236-L245).

This does not have to be  bug, but could also be a design choice. I don't really know which one it is (a bug or a design choice).. so we gotta make changes that are ideally backward compatible. My suggestion above was basically to use `stride` only for other layers and create a new arg in the Stacked2dCore to set the `input_stride` which could be used for the first layer. With these two args you can set them however you want in the RotationEquivarientCore when calling `super().__init__()`.

cblessing24 (approved):

cblessing24:
Looks good to me. Could you take one final look and if appropriate merge please @mohammadbashiri ?

mohammadbashiri (approved):

PR #187, opened 2023-02-14
Added global DataPoint and OutputPoint

Mvystrcilova:
Global `DataPoint` and `OutputPoint` were added in `TransformDataset` and `FileTreeDatasetBase` respectively to make them serializable for pickle. This was necessary to be able to use these datasets and set `num_workers` > 0 in `torch.utils.data.DataLoader`.

### DIFF
--- a/neuralpredictors/data/datasets/base.py
+++ b/neuralpredictors/data/datasets/base.py
@@ -90,7 +90,8 @@ class TransformDataset(Dataset):
             # this version IS serializable in pickle
             self.data_point = default_image_datapoint
         else:
-            # this version is NOT - you cannot use this with a dataloader with num_workers > 1
+            # DataPoint needs to be available for pickle in the global context to be serializable
+            global DataPoint
             self.data_point = namedtuple("DataPoint", data_keys)
 
     def transform(self, x, exclude=None):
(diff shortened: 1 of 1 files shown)


neuralpredictors/data/datasets/base.py
```diff
@@ -309,6 +310,9 @@ def __init__(self, dirname, *data_keys, transforms=None, use_cache=True, output_
         self.rename_output = bool(output_rename)
         self._output_rename = output_rename
         renamed_keys = [output_rename.get(k, k) for k in data_keys]
+        if output_rename:
+            # Output Point needs to be available for pickle in the global context 
+            global OutputPoint
```
mohammadbashiri, line 315: Does this work? I thought that the named tuple itself should be available in the global context. Have you tested it?
Also, If `output_rename` is `None` then `self.data_point` is used (next line) which is also a named tuple and should again be available in the global context. Or am I missing something?

mohammadbashiri (changes requested):

Mvystrcilova:
It is not deprecated I think, I just never actually fixed it. I would say it’s still worth fixing.

pollytur:
PR closed as it was a long time ago. A new PR with the same idea would be great

PR #192, opened 2023-03-10
Add everything for zero-inflated models

kklurz:
- losses
- encoders
- readouts
- fix of modulator


mohammadbashiri (commented):
Thanks @kklurz. Is this for the metric project release? and have you already tested if the code for the other project runs without errors using these modules from `neuralpredictors`? - just wanna make sure this is somehow tested because I don't think I can go though it in great detail.

kklurz:
Yes, this is for the metrics project release. And yes, I tested the code: The losses and models behave in the same way and give the same results as in the neural metrics code. I just restructured and rewrote the code in a way that I think is more general and better to use. Of you don't agree with the restructuring, we can discuss that of course.

mohammadbashiri:
> Yes, this is for the metrics project release. And yes, I tested the code: The losses and models behave in the same way and give the same results as in the neural metrics code. I just restructured and rewrote the code in a way that I think is more general and better to use. Of you don't agree with the restructuring, we can discuss that of course.

neuralpredictors/layers/modulators/mlp.py
```diff
@@ -26,7 +26,11 @@ def __init__(
             self.modulator_networks.append(nn.Sequential(*feat))
 
     def regularizer(self):
-        return self.linear.weight.abs().mean()
+        reg = 0
+        for mod in self.modulator_networks:
+            # regularize the weights of the last linear layer
+            reg += mod[-2].weight.abs().mean()
```
mohammadbashiri, line 32: This one seems to make an assumption about the modulator network, right? - that the second last layer is a linear layer, and what is the last one? a nonlinearity? Is there a more general way to find the last linear layer in the network? It might be that the modulator is constructed without a final nonlinearity, no?

kklurz: The way that the `__init__` method is constructed at the moment, the last linear layer will always be the index `-2`. This would break, of course, if the `__init__` was changed in a future PR without taking care of changing the rest of the class. If we want to exclude this possible future bug, how would you implement finding the last linear layer? Something like this?
```
reg = 0
for mod in self.modulator_networks:
    # find the last layer
    for l in range len(mod):
        if isinstance(mod[-l], nn.Linear):
            last_linear_layer = mod[-l]
            break
            
    # regularize the weights of the last linear layer
    reg += last_linear_layer.weight.abs().mean()
```

mohammadbashiri: Yes, this looks good - except that the `()` are missing for range. 
I also took a look at the `__init__` method and you are right that it is very specific in how the model is constructed and this way of checking for the last linear layer is probably an overkill. So up to you. I would suggest the second option so that we have a good example for the future and also if someone changed it later then it does the expected thing (i.e. regularize the last linear layer).

PR #193, opened 2023-03-16
Simplify Independent bn bias

KonstantinWilleke:
and remove SE2d core.


MaxFBurg:
`if independent_bn_bias:` results in this branch of the if else: https://github.com/sinzlab/neuralpredictors/pull/193/files#diff-41911c485f9635deecaf442f001fcde8fdf8606023624d9844a14642c5820a41L212 , and the initialized `self.batchnorm_layer_cls(hidden_channels, momentum=self.momentum)` does not take `self.bias` or `self.batch_norm_scale` or `final_batchnorm_scale`, hence if we set these to other than the default values, they will be ignored. The default values are all `True`. So if any of these values is not default, i.e. `False` , they will be ignored. Hence, we should check if any value deviates from default (`if not bias or not batch_norm_scale or not final_batchnorm_scale`,) and warn in this case:
```
                warnings.warn(
                    "Setting `independent_bn_bias=True` leads to `bias` or `batch_norm_scale` or "
                    "`final_batchnorm_scale` being ignored."
                )
```

KonstantinWilleke:
Yes you're right, thanks for pointing that out! I changed it so that there are `not`s  everywhere, and I improved the warning a bit by making it more explicit. Feel free to merge now @MaxFBurg!

MaxFBurg (approved):

MaxFBurg:
Great, thank you very much @KonstantinWilleke ! I'm happy with merging now, just the first test did not pass yet

PR #200, opened 2023-04-14
Fix bug where concentration and rate were flipped

kklurz:


suhasshrinivasan (approved):
This was a very nasty bug! Thanks a lot @kklurz for finding and fixing it. Saved us many hours!

PR #202, opened 2023-04-22
For sensorium 2023

pollytur:
This PR implements some of the video data transformations 
* changing channels order
* cutting nans at the end of files (artefact from nexport as it requires videos to be of the same lengths while they are not)
* adding behaviour as channels for video data (minor changes with keys in transformation functions)
and adds GRU module implemented by Sinz et all at 2018 (code is mostly borrowed from there with references)

### DIFF
--- a/neuralpredictors/data/transforms.py
+++ b/neuralpredictors/data/transforms.py
@@ -1,5 +1,6 @@
 from collections import namedtuple
 from functools import partial
+
 try:
     from collections import Iterable
 except:
--- a/neuralpredictors/data/transforms.py
+++ b/neuralpredictors/data/transforms.py
@@ -1,9 +1,9 @@
 from collections import Iterable, namedtuple
+from functools import partial
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


neuralpredictors/data/transforms.py
```diff
@@ -400,26 +435,23 @@ class AddBehaviorAsChannels(MovieTransform, StaticTransform, Invertible):
         - behavior
     """
 
-    def __init__(self):
+    def __init__(self, key):
```
pollytur, line 438: make default as image to make it back compatible

neuralpredictors/data/transforms.py
```diff
@@ -439,16 +471,13 @@ class AddPupilCenterAsChannels(MovieTransform, StaticTransform, Invertible):
         - pupil center
     """
 
-    def __init__(self):
+    def __init__(self, key):
```
pollytur, line 474: here same

mohammadbashiri (approved):

PR #204, opened 2023-04-25
Fixing output shape of 3d cores + more hyperparameter options

Mvystrcilova:
- removing get_core_output_shape function from both 3d cores because it does not work with padding
- allowing different number of hidden channels in 3d cores
- allowing different dilations for hidden layers in 3d cores


Mvystrcilova:
Guys, could someone review this (assign a reviewer - I don't have the rights) and accept this request. Without it, the notebooks for sensorium do not work. Thanks!

pollytur (approved):
look good for me

kklurz (approved):

PR #207, opened 2023-05-23
Added logic to support warm-up and all pytorch schedulers 

vdobrev1:
Currently, the early_stopping.py file only supports the [ReduceLROnPlateau ](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ReduceLROnPlateau.html)scheduler from the pytroch schedulers, as it is the only one that accepts input in its step method. In order to use any of the other pytorch schedulers the step method must be called without input. This push request contains code allowing the rest of the schedulers to be used and adds the option to use a warm-up scheduler.

- Added logic to check if a provided scheduler is ReduceLROnPlateau in which case the step method is called with the current objective.
- Added logic to support a tuple of schedulers as input where the first one is treated as a warm-up scheduler and is called a number of times defined by the user, afterwards only the second one is called.
- Added examples in the docstring of valid scheduler tuples and an example of how to define a warm-up scheduler

Note: The warm-up is implemented using a tuple and not the [SequentialLR ](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.SequentialLR.html#torch.optim.lr_scheduler.SequentialLR)scheduler from pytorch since it does not work with ReduceLROnPlateau. As before the schedulers must be defined in the main training loop.


pollytur (approved):
would need modifications to have more schedulers later but good for now

PR #209, opened 2023-06-23
handle case when there is no NaN value in CutVideos

bryanlimy:
- `np.argwhere(np.isnan(loc))[:, self.frame_axis[k]].min()` would cause an error when there is no NaN values in `loc` (calling `.min()` on an empty array).
- added a quick fix to check if `np.argwhere(np.isnan(loc))[:, self.frame_axis[k]]` is empty, if so, set `idx` to the length of `loc`, otherwise return `min()` value.

### DIFF
--- a/neuralpredictors/data/transforms.py
+++ b/neuralpredictors/data/transforms.py
@@ -653,7 +653,8 @@ class CutVideos(MovieTransform):
             idx = []
             for k in x._fields:
                 loc = getattr(x, k)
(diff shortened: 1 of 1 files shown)


pollytur:
Hi, thanks for your suggestion!

PS If you tried to delete nans - just make sure, if you want or do not want to use normalisation (as videos are of a different length but we normalise them together in dataset augemntetations, removing nans might cause troubles with normalisation)

PR #210, opened 2023-08-17
fix that cannot correctly set bn layer channels

wurining:
When setting the `hidden_channels` as a integer, function `add_bn_layer` was passed a error parameter which caused TypeError as below

```
Traceback (most recent call last):
  File "/***/sensorium_2023/main.py", line 143, in <module>
    full_3d_model = make_video_model(
  File "/***/sensorium_2023/sensorium_2023/sensorium/models/make_model.py", line 141, in make_video_model
    core = Basic3dCore(**core_dict)
  File "/***/conda_envs/envs/sensorium/lib/python3.9/site-packages/neuralpredictors/layers/cores/conv3d.py", line 169, in __init__
    self.add_bn_layer(layer=layer, hidden_channels=hidden_channels[0])
TypeError: 'int' object is not subscriptable
```


MaxFBurg (approved):

MaxFBurg:
Thank you very much for the example and fix @wurining - we highly appreciate your contribution!

PR #215, opened 2023-12-14
Fixes bug that would add a redundant scale layer to batchnorm layer with affine=True

MaxFBurg:
When creating a model with these params

```
batch_norm=True,
independent_bn_bias=False,
bias=True,
batch_norm_scale=True,
```

we would see redundant scale parameters (because in `affine=True` in `BatchNorm1d`:

```
      (norm): RotationEquivariantBatchNorm2D(
        (batch_norm): BatchNorm1d(8, eps=1e-05, momentum=0.1, affine=True, track_running_stats=True)
      )
      (scale): RotationEquivariantScale2DLayer()
```

Code to reproduce:

```
from neuralpredictors.layers.cores import RotationEquivariant2dCore
core = RotationEquivariant2dCore(
    pad_input=False,
    stack=-1,
    layers=4,
    hidden_channels=8,
    num_rotations=8,
    input_kern=13,
    hidden_kern=5,
    gamma_input=6,
    batch_norm=True,
    independent_bn_bias=False,
    bias=True,
    batch_norm_scale=True,
    final_batchnorm_scale=False,
    input_channels=1,
)
print(core)
```

### DIFF
--- a/neuralpredictors/layers/cores/conv2d.py
+++ b/neuralpredictors/layers/cores/conv2d.py
@@ -6,6 +6,7 @@ try:
     from collections import Iterable
 except:
     from collections.abc import Iterable
+
 from functools import partial
 
 import torch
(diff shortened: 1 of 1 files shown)


pollytur:
I am not sure if its a bug, based on the [docstring](https://github.com/sinzlab/neuralpredictors/blob/e56c7a790a8b82baf8a5909d8041e2b52c386363/neuralpredictors/layers/cores/conv2d.py#L93C13-L93C82) it might be not 
`batch_norm_scale: If True, a scaling factor after BN will be learned.`

Issue #175, opened 2022-08-12
Wrong order of dimensions for readout in_shape


KonstantinWilleke:
When we pass the shape of the core output tensor to the readout, we consistently have the wrong order of `height` and `width`.

```
c, w, h = in_shape
```
(from https://github.com/sinzlab/neuralpredictors/blob/711ffeeead468aaca0c0b9dac0cc66419f7faa83/neuralpredictors/layers/readouts/gaussian.py#L64)

it's not a bug per se, because it is consistent internally. But would be clean to change it for all readouts.

MaxFBurg:
Should be same as what is commonly used in pytorch, i.e. (batch, channel, height, width), right?

pollytur:
Solved in #225

PR #225, opened 2024-03-07
Fix/shape dimension order

MaxFBurg:
Addresses https://github.com/sinzlab/neuralpredictors/issues/175


pollytur (approved):
Sequence might cause problems for python below 3.6 but I believe that's neglectable in 2024:)

Issue #218, opened 2024-02-13
[improvement] Factorised readouts - positivity constraint on masks only 


pollytur:
This is important to make the code for older papers reproducible in terms of qualitative findings 

The masks should have a restriction to be positive independently of the feature readouts like [here](https://github.com/aecker/cnn-sys-ident/blob/master/cnn_sys_ident/architectures/readouts.py#L37) it is independent of feature weights positivity constraint (which is [here](https://github.com/aecker/cnn-sys-ident/blob/master/cnn_sys_ident/architectures/readouts.py#L45)), while currently in neuropredictors its either everything is positive or nothing ([here](https://github.com/sinzlab/neuralpredictors/blob/main/neuralpredictors/layers/readouts/factorized.py#L20) - it restricts both weights and masks being positive)

pollytur:
Solved in #232

PR #232, opened 2024-03-08
Added the restriction on Factorised masks to be positive using absolute value (as in ecker 2018)

pollytur:
#218

### DIFF
--- a/neuralpredictors/layers/readouts/factorized.py
+++ b/neuralpredictors/layers/readouts/factorized.py
@@ -36,7 +36,7 @@ class FullFactorized2d(Readout):
             in_shape: batch, channels, height, width (batch could be arbitrary)
             outdims: number of neurons to predict
             bias: if True, bias is used
-            normalize: if True, normalizes mask using by l2 norm
+            normalize: if True, normalizes the spatial mask using l2 norm
             init_noise: the std for readout  initialisation
             constrain_pos: if True, negative values in the neuronal features are turned into 0
             positive_weights: if True, negative values in the mask are turned into 0
--- a/neuralpredictors/layers/readouts/factorized.py
+++ b/neuralpredictors/layers/readouts/factorized.py
@@ -38,9 +38,9 @@ class FullFactorized2d(Readout):
             bias: if True, bias is used
             normalize: if True, normalizes the spatial mask using l2 norm
             init_noise: the std for readout  initialisation
-            constrain_pos: if True, negative values in the neuronal features are turned into 0
-            positive_weights: if True, negative values in the mask are turned into 0
-            positive_spatial: if True, mask values are restricted to be positive by taking absolute values
+            constrain_pos: if True, negative values in the spatial mask and feature readout are clamped to 0
+            positive_weights: if True, negative values in the feature readout are turned into 0
+            positive_spatial: if True, spatial readout mask values are restricted to be positive by taking the absolute values
             shared_features: if True, uses a copy of the features from somewhere else
             mean_activity: the mean for readout  initialisation
             spatial_and_feature_reg_weight: lagrange multiplier (constant) for L1 penalty,
--- a/neuralpredictors/layers/readouts/factorized.py
+++ b/neuralpredictors/layers/readouts/factorized.py
@@ -23,7 +23,7 @@ class FullFactorized2d(Readout):
         init_noise=1e-3,
         constrain_pos=False,
         positive_weights=False,
-        positive_spatial=True,
+        positive_spatial=False,
         shared_features=None,
         mean_activity=None,
         spatial_and_feature_reg_weight=None,
--- a/neuralpredictors/layers/readouts/factorized.py
+++ b/neuralpredictors/layers/readouts/factorized.py
@@ -45,7 +45,7 @@ class FullFactorized2d(Readout):
             mean_activity: the mean for readout  initialisation
             spatial_and_feature_reg_weight: lagrange multiplier (constant) for L1 penalty,
                 the bigger the number, the stronger the penalty
-            gamma_readout: depricated, use feature_reg_weight instead
+            gamma_readout: depricated, use spatial_and_feature_reg_weight instead
             **kwargs:
         """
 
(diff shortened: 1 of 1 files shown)


neuralpredictors/layers/readouts/factorized.py
```diff
@@ -84,6 +113,8 @@ def normalized_spatial(self):
             weight = self.spatial
         if self.constrain_pos:
             weight.data.clamp_min_(0)
+        elif self.positive_spatial:
+            weight = torch.abs(weight)
```
MaxFBurg, line 117: I think it would be more intuitive if `positive_spatial` would be implemented in the same way as `constrain_pos`. The overall result of the trained network parameters should be the same.

```suggestion
        if self.constrain_pos or self.positive_spatial:
            weight.data.clamp_min_(0)
```

neuralpredictors/layers/readouts/factorized.py
```diff
@@ -50,7 +79,7 @@ def __init__(
         else:
             self.register_parameter("bias", None)
 
-        self.initialize(mean_activity)
+        self.initialize()
```
MaxFBurg, line 82:
```suggestion
        self.initialize(mean_activity)
```
I think we should not change the default behavior of code if we do not have too, as this would break models we already have trained / result in reproducibility issues.

neuralpredictors/layers/readouts/factorized.py
```diff
@@ -19,12 +23,31 @@ def __init__(
         init_noise=1e-3,
         constrain_pos=False,
         positive_weights=False,
+        positive_spatial=True,
```
MaxFBurg, line 26:
```suggestion
        positive_spatial=False,
```
Reasoning: I think we should not change the default behavior of code if we do not have too, as this would break models we already have trained / result in reproducibility issues.

neuralpredictors/layers/readouts/factorized.py
```diff
@@ -19,12 +23,31 @@ def __init__(
...
+        gamma_readout=None,
         **kwargs,
     ):
+        """
+
+        Args:
+            in_shape: batch, channels, height, width (batch could be arbitrary)
+            outdims: number of neurons to predict
+            bias: if True, bias is used
+            normalize: if True, normalizes mask using by l2 norm
```
MaxFBurg, line 39: I think - if I am not mistaken - this only affects the spatial mask, so I would suggest to be precise here for users that do not know the code yet in detail
```suggestion
            normalize: if True, normalizes the spatial mask using l2 norm
```

neuralpredictors/layers/readouts/factorized.py
```diff
@@ -19,12 +23,31 @@ def __init__(
...
+            init_noise: the std for readout  initialisation
+            constrain_pos: if True, negative values in the neuronal features are turned into 0
+            positive_weights: if True, negative values in the mask are turned into 0
+            positive_spatial: if True, mask values are restricted to be positive by taking absolute values
+            shared_features: if True, uses a copy of the features from somewhere else
+            mean_activity: the mean for readout  initialisation
+            spatial_and_feature_reg_weight: lagrange multiplier (constant) for L1 penalty,
+                the bigger the number, the stronger the penalty
+            gamma_readout: depricated, use feature_reg_weight instead
```
MaxFBurg, line 48:
```suggestion
            gamma_readout: depricated, use spatial_and_feature_reg_weight instead
```
We could think of additionally raising a user warning by `warnings.warn`

neuralpredictors/layers/readouts/factorized.py
```diff
@@ -33,6 +56,12 @@ def __init__(
         self.outdims = outdims
         self.positive_weights = positive_weights
         self.constrain_pos = constrain_pos
+        self.positive_spatial = positive_spatial
+        if positive_spatial and constrain_pos:
+            logger.warning(
```
MaxFBurg, line 61: Fabi or Edgar once told us in a code review sessions, that any warning that was triggered because the user did something that they should not, people usually would use `warnings.warn`; whereas warnings the user cannot do anything about would be triggered by `logger.warning`. So in this case it would mean to rather use `warnings.warn`

neuralpredictors/layers/readouts/factorized.py
```diff
@@ -19,12 +23,31 @@ def __init__(
...
+
+        Args:
+            in_shape: batch, channels, height, width (batch could be arbitrary)
+            outdims: number of neurons to predict
+            bias: if True, bias is used
+            normalize: if True, normalizes mask using by l2 norm
+            init_noise: the std for readout  initialisation
+            constrain_pos: if True, negative values in the neuronal features are turned into 0
+            positive_weights: if True, negative values in the mask are turned into 0
+            positive_spatial: if True, mask values are restricted to be positive by taking absolute values
```
MaxFBurg, line 43:
```suggestion
            constrain_pos: if True, negative values in the spatial mask and feature readout are clamped to 0
            positive_weights: if True, negative values in the feature readout are turned into 0
            positive_spatial: if True, spatial readout mask values are restricted to be positive clamping them to 0
```

MaxFBurg (changes requested):

neuralpredictors/layers/readouts/factorized.py
MaxFBurg: In tensorflow 1.x there was no clamp_min afaik, hence we used abs

pollytur: I actually would like to keep it as in the original code [here](https://github.com/aecker/cnn-sys-ident/blob/master/cnn_sys_ident/architectures/readouts.py#L37) 

Because clamping might influence the gradient descent (because its zeros) and this might influence the performance

neuralpredictors/layers/readouts/factorized.py
pollytur: I thing this might be a different PR probably because we might change it consistently everywhere in readouts (not only here)

MaxFBurg (approved):

PR #233, opened 2024-03-08
Tutorial for 2d cores added

neuronmorph:
Tutorial provides examples on how to use different 2d cores for sensorium 2022


pollytur:
Based on discussion with Konsti

- small configs for each core type, highlighting the differences
- `core_config` not `model_config`
- change a bit the description for the SE core 
- should loading and not loading a ppretrain core for transfer learning

pollytur:
@KonstantinWilleke 
The installation is there 
image


I actually would disagree with the explanation of sensorium dependencies, as it's not the goal of this tutorial- the goal is to show the cores. And I expect for some people to be very distracted by this from the beginning.

Using random inputs is actually a great idea and would make things easier and more disentangled. This way we do not need either installation or explanations.

pollytur (approved):
Could you please update it with the random input for the cores as Konsti suggested (removing the dataset creation and sensorium installation) and then we could merge this great example?

pollytur (changes requested):
Could you please update it with the random input for the cores as Konsti suggested (removing the dataset creation and sensorium installation) and then we could merge this great example?

neuronmorph:
Please have a look. I removed the Sensorium 2022 dependencies.

pollytur:
Great, thanks a lot!

pollytur (approved):

KonstantinWilleke (approved):

KonstantinWilleke:
Great work @neuronmorph! Thanks a lot. You can merge the PR - all that is left to do is to update the branch. You can click the button "update branch" and then feel free to merge.

Issue #229, opened 2024-03-08
[Bug fix?] Shift replaced if shifter defined


pollytur:
[Here](https://github.com/sinzlab/neuralpredictors/blob/main/neuralpredictors/layers/encoders/firing_rate.py#L73) the original shift would be replaced is the shifter is defined. This is probably not the expected behaviour

pollytur:
solved in #234

PR #234, opened 2024-03-08
fix_shift_replacement_if_defined_2d_encoder

pollytur:
#229


MaxFBurg:
I agree. As I was not involved in developing the shifter so much, it would be good if @KonstantinWilleke, @kklurz or @mohammadbashiri could confirm that

mohammadbashiri:
Thanks @pollytur for the PR. The change looks good and makes sense.

MaxFBurg (approved):

PR #235, opened 2024-03-08
Added readout basics tutorial notebook

fededagos:
First version of a readout tutorial notebook made during the Hackathon. Feel free to modify and edit it as needed.


pollytur (changes requested):
Thanks a lot for the great PR! I really like how detailed it is and the parameter comparison in the end.

small comment to improve
* the import of `nnfabric` happens at the very first cell, while it installation is much later. Also, the functions, used from nnfabric are more like 1-10 line wrappers for the internal infrastructure, so I would say ideally this could be just removed at all from the examples in these tutorials. Feel free to copy paste them from [here](https://github.com/sinzlab/nnfabrik/blob/5b6e7379cb5724a787cdd482ee987b8bc0dfacf3/nnfabrik/utility/nn_helpers.py#L39), [here](https://github.com/sinzlab/nnfabrik/blob/5b6e7379cb5724a787cdd482ee987b8bc0dfacf3/nnfabrik/utility/nn_helpers.py#L53), and [here](https://github.com/sinzlab/nnfabrik/blob/5b6e7379cb5724a787cdd482ee987b8bc0dfacf3/nnfabrik/builder.py#L87)

pollytur:
@KonstantinWilleke What do you think about using sensorium data here? The first explanation is without the data, but I feel like for readouts its more needed to have a real data example on how the shapes are found out, etc.

Issue #240, opened 2024-05-28
[bug] Batch norm layers for 3d core


pollytur:
After hackathon adjustments, `add_bn_layer` is inherited from [ConvCore](https://github.com/sinzlab/neuralpredictors/blob/main/neuralpredictors/layers/cores/base.py#L74) but the function there does not take `hidden_channels` while there are needed for the 3d core ([here](https://github.com/sinzlab/neuralpredictors/blob/main/neuralpredictors/layers/cores/conv3d.py#L163))

Also, would be great, instead of throwing errors to make things lists [here](https://github.com/sinzlab/neuralpredictors/blob/main/neuralpredictors/layers/cores/base.py#L78) and add a warning. Otherwise, most of the public repos are not back compatible and its easy to make them

pollytur:
solved in #241

PR #241, opened 2024-05-28
batch norms fix for 3d core

pollytur:
This one basically fixes this issue #240


yongrong-qiu:
Good idea! OK with me

PR #242, opened 2024-06-06
Factorial fix

pollytur:
math.factorial requires int as input - explicit typing was needed

Otherwise with python 3.12, numpy 1.26.4 and torch 2.2.2 the error occurred

image


yongrong-qiu:
Ok

yongrong-qiu (approved):

PR #245, opened 2025-07-09
Perspective module

NathanSoeding:
simulates light rays emitted from the screen received by the mices retina

### DIFF
--- a/neuralpredictors/layers/encoders/firing_rate.py
+++ b/neuralpredictors/layers/encoders/firing_rate.py
@@ -68,8 +68,12 @@ class FiringRateEncoder(Encoder):
         x = inputs
 
         if self.perspective:
+            if self.shifter:
+                raise ValueError("both perspective and shifter cannot be present together, only one should be chosen")
+            
             if pupil_center is None:
                 raise ValueError("pupil_center is not given")
+            
             x = self.perspective[data_key](x, pupil_center)
 
         x = self.core(x)
--- a/neuralpredictors/layers/encoders/firing_rate.py
+++ b/neuralpredictors/layers/encoders/firing_rate.py
@@ -13,6 +13,7 @@ class FiringRateEncoder(Encoder):
         core,
         readout,
         *,
+        perspective=None,
         shifter=None,
         modulator=None,
         elu_offset=0.0,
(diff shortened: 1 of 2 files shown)


neuralpredictors/layers/perspective/mlp.py
```diff
@@ -0,0 +1,272 @@
+import torch
+from torch import nn
+
+
+def angles_to_rmat3d(angles):
```
pollytur, line 10: could you please add some small docstring here explaining conceptually what this function is doing

neuralpredictors/layers/encoders/firing_rate.py
```diff
@@ -63,7 +65,14 @@ def forward(
         detach_core=False,
         **kwargs
     ):
-        x = self.core(inputs)
+        x = inputs
+
+        if self.perspective:
+            if pupil_center is None:
```
pollytur, line 74:
```suggestion
            if self.shifter:
                raise ValueError("both perspective and shifter cannot be present together, only one should be chosen")
            if pupil_center is None:
```

neuralpredictors/layers/perspective/mlp.py
```diff
@@ -0,0 +1,272 @@
+import torch
```
pollytur, line 7:
```suggestion
'''
code is adjusted from the Wang et al 2025 "Foundation model of neural activity predicts response to new stimulus types" implementation
Specifically
* https://github.com/cajal/fnn/blob/main/fnn/model/pixels.py
* https://github.com/cajal/fnn/blob/main/fnn/model/perspectives.py
'''
import torch
```

neuralpredictors/layers/perspective/mlp.py
pollytur, line 1: @NathanSoeding could you please add minimal docstrings everywhere in this file, basically what is the conceptual idea of each class and same for some of the functions
Thanks a lot in advance!

pollytur (changes requested):
Thanks a lot for the PR

Please take a look at a couple of suggestions and please add a bit of the docstrings to the `mlp.py`
Otherwise, looks good for me

pollytur (approved):
awesome, thanks a lot!

PR #247, opened 2026-06-29
shifter init fix

NathanSoeding:
shifter initialization function was never called and it checked if parameters are instances of nn.Linear which is always wrong because they are tensors.

### DIFF
--- a/neuralpredictors/layers/shifters/mlp.py
+++ b/neuralpredictors/layers/shifters/mlp.py
@@ -31,13 +31,18 @@ class MLP(Shifter):
         feat.extend([nn.Linear(prev_output, 2, bias=bias), nn.Tanh()])
         self.mlp = nn.Sequential(*feat)
 
+        self.initialize()
+
     def regularizer(self):
         return 0
 
     def initialize(self):
-        for linear_layer in [p for p in self.parameters() if isinstance(p, nn.Linear)]:
-            xavier_normal(linear_layer.weight)
-
+        for layer in self.mlp:
+            if isinstance(layer, nn.Linear):
+                xavier_normal_(layer.weight)
+                if layer.bias is not None:
+                    nn.init.zeros_(layer.bias)
+        
     def forward(self, pupil_center, trial_idx=None):
         if trial_idx is not None:
             pupil_center = torch.cat((pupil_center, trial_idx), dim=1)
@@ -73,9 +78,7 @@ class MLPShifter(ModuleDict):
             self.add_module(k, MLP(input_channels, hidden_channels_shifter, shift_layers, bias))
 
     def initialize(self, **kwargs):
-        logger.info("Ignoring input {} when initializing {}".format(repr(kwargs), self.__class__.__name__))
-        for linear_layer in [p for p in self.parameters() if isinstance(p, nn.Linear)]:
-            xavier_normal(linear_layer.weight)
+        pass
 
     def regularizer(self, data_key):
         return self[data_key].regularizer() * self.gamma_shifter
--- a/neuralpredictors/layers/shifters/mlp.py
+++ b/neuralpredictors/layers/shifters/mlp.py
@@ -78,5 +78,9 @@ class MLPShifter(ModuleDict):
         for k in data_keys:
             self.add_module(k, MLP(input_channels, hidden_channels_shifter, shift_layers, bias))
 
+    def initialize(self, **kwargs):
+         for k in self:
+             self[k].initialize()
+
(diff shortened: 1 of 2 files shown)


neuralpredictors/layers/shifters/mlp.py
```diff
@@ -73,9 +78,7 @@ def __init__(
             self.add_module(k, MLP(input_channels, hidden_channels_shifter, shift_layers, bias))
 
     def initialize(self, **kwargs):
-        logger.info("Ignoring input {} when initializing {}".format(repr(kwargs), self.__class__.__name__))
-        for linear_layer in [p for p in self.parameters() if isinstance(p, nn.Linear)]:
-            xavier_normal(linear_layer.weight)
+        pass
```
pollytur: @NathanSoeding why do you pass it here? Its rather a silent failing / no initialization (default torch init, which is kaiming uniform initialization and not  xavier_normal) - could we maybe properly fix it instead or do I miss sth?

pollytur: i guess with the current changes this function is redundant - why not to delete at all?

neuralpredictors/layers/shifters/mlp.py
```diff
@@ -31,13 +31,18 @@ def __init__(self, input_features=2, hidden_channels=10, shift_layers=1, bias=Tr
...
 
     def initialize(self):
-        for linear_layer in [p for p in self.parameters() if isinstance(p, nn.Linear)]:
-            xavier_normal(linear_layer.weight)
-
+        for layer in self.mlp:
+            if isinstance(layer, nn.Linear):
+                xavier_normal_(layer.weight)
+                if layer.bias is not None:
+                    nn.init.zeros_(layer.bias)
```
pollytur: And ideally you want to do sth like this (to protect from user mistakes if the gradients are started to track earlier somehow) - [torch implementations do it as well](https://github.com/pytorch/pytorch/blob/39f94619ec3bf827567716e6c9b5003ef20036f4/torch/nn/init.py#L50)
```suggestion
        with torch.no_grad():
            for layer in self.mlp:
                if isinstance(layer, nn.Linear):
                    xavier_normal_(layer.weight)
                    if layer.bias is not None:
                        nn.init.zeros_(layer.bias)
```

neuralpredictors/layers/shifters/mlp.py
NathanSoeding: I passed because the initialization happens already in the creation of each shifter individually so this init should be redundant. Yes before it just used standard pytorch initialization (kaiming).

neuralpredictors/layers/shifters/mlp.py
NathanSoeding: Ok makes sense

pollytur (changes requested):
@NathanSoeding thanks a lot for the PR!

see the comments to address

Ideally please add a few unit tests to check that initialization now works correctly for both `MLP(Shifter)` and `class MLPShifter(ModuleDict)`  to make sure we are not missing any other bugs :)

neuralpredictors/layers/shifters/mlp.py
pollytur: yes, but why not to delete the function completely then?

neuralpredictors/layers/shifters/mlp.py
pollytur: then please fix and let me know when to take a look again :)

pollytur:
@NathanSoeding any updates on it? It would be great if we could finalise it soon :)

NathanSoeding:
I committed the changes

test/test_shifter_init.py
```diff
@@ -0,0 +1,12 @@
+import torch
+from torch import nn
+from unittest.mock import patch
```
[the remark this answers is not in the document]
pollytur: @NathanSoeding please fix this :)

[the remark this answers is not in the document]
pollytur: and also maybe add a sweep that you sent several values of the MLP layers and not just hardcoded 3 - in experanto there should be examples how to parametric tests with pytest like here https://github.com/sensorium-competition/experanto/blob/main/tests/test_sequence_interpolator.py#L14-L16

If pytest is not a part of neural predictors - ignore this comment and just remove the unused inputs

Have you run the tests locally?

NathanSoeding:
@pollytur I removed unused library imports and added added the module dict initialization function which calls the individual shifter initialize() functions

pollytur (approved):

